PR #221, opened 2012-09-27
Don't expose the existence of a user when resetting

aripollak:
If an anonymous visitor submits a "reset password" request, they get a different response based on whether or not the email exists in the database. This goes against best security practices:

https://wiki.mozilla.org/WebAppSec/Secure_Coding_Guidelines#Password_Reset_Functions

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -9,11 +9,8 @@ class Clearance::PasswordsController < ApplicationController
     if user = find_user_for_create
       user.forgot_password!
       ::ClearanceMailer.change_password(user).deliver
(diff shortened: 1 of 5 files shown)


rolftimmermans:
Although this is a good idea from a security perspective, the user experience is pretty poor. Users frequently forget which email address they used, and this makes it pretty much impossible to reset their password if they are uncertain about it.

Is there interest in making this configurable? If so, we can submit a pull request for it.

derekprior:
I feel pretty strongly thiat this is an important security feature. I think a better way to improve messaging would be to deliver an email to the provided address which says something like "you tried to reset your password but we don't have a user with this login"

This way we keep the security but have some indication to the user of what happened.

PR #398, opened 2014-02-14
Backfill remember_token for existing user records

### LINKED ISSUE #396: Add and backfill `remember_token` for existing records in existing app.
If an application already has a model like `User` and they add Clearance, we need to generate `remember_token`s for those records. The best approach is to do this via migration. Refer to #343 for history.

derekprior:
The clearance migration to add clearance fields to an existing users table now
also ensures each user has a generated remember_token.

Closes #396

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 Thank you to all the [contributors](https://github.com/thoughtbot/clearance/graphs/contributors)!
 
+New on MASTER
+* Installing Clearance with an existing User model will now add remember tokens
+  to all user records.
+
 New for 1.2.1 (March 6, 2014):
 * Query string is now included in the redirect URL when Clearance redirects to a
   previously stored URL.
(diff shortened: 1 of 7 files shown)


jjb:
i recently had to do this very thing, didn't know about this migration, and rolled my own. my initial implementation was very similar to what's in this PR, and it took many minutes for my 200+k users. Then I switched to the approach below, which takes about 10 seconds.

it consists of this:

```ruby
ActiveRecord::Base.connection.execute <<-SQL
  update users set remember_token = md5(random()::text);
SQL
```

Just noting it here in case someone reading this PR in the future will find it useful.

derekprior:
There was some discussion about why we didn't go that direction in an outdated diff on this PR: https://github.com/thoughtbot/clearance/pull/398#discussion-diff-9761120

Additionally, I am not a crypto expert, but you may want to use `gen_random_bytes` from pgcrypto instead.

PR #413, opened 2014-03-11
Clearance controllers publicize themselves as just that

zamith:
- This allows performing actions on controllers conditionally, such as
  [checking authorization using cancan](https://github.com/ryanb/cancan/wiki/Ensure-Authorization#wiki-conditionally-check-authorization) for all controllers except for the
  ones in clearance.
- Adds a BaseController

The desired behaviour would be to accomplish something like this:

```
class ApplicationController < ActionController::Base
  check_authorization unless: :clearance_controller?
end
```


zamith:
@derekprior I just found myself wanting to do this again, and @jferris comment seems like a good way to go about it. Should this be documented somewhere?

zamith:
@derekprior Actually @jferris' solution had an issue in development, since the changes were applied in an initializer, once some classes were reloaded, they would stop working. I was able to solve it like this:

```ruby
def setup_clearance_base_controller
  Clearance::BaseController.class_eval do
    layout 'clearance'
    skip_before_filter :authorize
  end
end

if Rails.configuration.cache_classes
  setup_clearance_base_controller
else
  ActionDispatch::Reloader.to_prepare do
    setup_clearance_base_controller
  end
end
```

PR #500, opened 2014-12-30
Dont set cookie header if not present

### LINKED ISSUE #338: Clearance overwrites secure-only cookies on HTTP requests
1. Configure a site with `force_ssl` in the ApplicationController
2. Configure clearance with `config.secure_cookie = true`
3. Log in to the site
4. Navigate away to some other site.
5. Get back to the site by typing in `www.domain.com` (without https)

You are not signed in. Your browser did not send your previous remember token cookie because it wasn't an HTTPS request. Clearance didn't see a remember token, so it set one to nil (secure only, again). You were then redirected to HTTPS, but future HTTPS requests send the nil token, so the user is not remembered.

I would expect clearance not to set the remember token at all if the request is not over HTTPS.

Here's the cURL output for the HTTP request:

```
➜  curl -I http://myglu.org
HTTP/1.1 301 Moved Permanently
Cache-Control: no-cache
Content-Type: text/html; charset=utf-8
Date: Tue, 13 Aug 2013 17:23:53 GMT
Location: https://myglu.org/
Set-Cookie: remember_token=; path=/; secure
Status: 301 Moved Permanently
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Rack-Cache: miss
X-Request-Id: d58eb7b3-8a13-401a-a1e7-6ce128f42906
X-Runtime: 0.029015
X-Ua-Compatible: chrome=1
X-Xss-Protection: 1; mode=block
Connection: keep-alive
```

Notice the `Set-Cookie` header.

arthurnn:
[fixes #338]

Trying to fix this problem on rubygems.org which uses clearance. The problem is well described in the issue.

### Solution
clearance always set the response cookie, even when is emtpy, so when doing a redirect, http->https, it would set an empty cookie, which is not what we want.
I cannot see any problem about not setting the cookie when it is empty. (please let me know if I miss something)

I hope CI passes, as I could not run tests locally ( see error: https://gist.github.com/arthurnn/e4bc23aecdc052870b0d )

review @derekprior

### DIFF
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -9,7 +9,9 @@ module Clearance
     end
 
     def add_cookie_to_headers(headers)
-      Rack::Utils.set_cookie_header!(headers, REMEMBER_TOKEN_COOKIE, cookie_value)
+      if cookie_value[:value].present?
+        Rack::Utils.set_cookie_header!(headers, REMEMBER_TOKEN_COOKIE, cookie_value)
+      end
     end
 
     def current_user
--- a/spec/clearance/session_spec.rb
+++ b/spec/clearance/session_spec.rb
@@ -320,7 +320,7 @@ describe Clearance::Session do
     headers = {}
     session = Clearance::Session.new(env_without_remember_token)
     session.add_cookie_to_headers headers
-    expect(headers).not_to set_cookie('remember_token')
+    expect(headers["Set-Cookie"]).to be nil
   end
 
   it 'signs out a user' do


derekprior:
@arthurnn any luck with the tests for this?

arthurnn:
sorry man.. will work onthose in the next hours

arthurnn:
@derekprior I actually found out that the spec to ensure this behaviour was there already, but the expectation on it was not right. So I just fixed, proving the point. (You can read the comment on commit 7779b12 to understand better what I mean)

derekprior:
looks like a really good find, thanks. I'm going to play with this a bit in some test apps to make sure I'm not missing something, but this looks good.

derekprior:
merged as 3a1244346524d9610105c2e5411bb759c6c05869. Thanks so much for taking a look at this. I'll be release 1.6.1 with this soon.

arthurnn:
thats awesome. this will solve a long time issue in rubygems.org.. 
:heart: thanks a lot for improving the commit message too.. looks great.

PR #505, opened 2015-01-06
Fix warnings generated by Clearance

derekprior:
* Instance variables should be defined in initializers before used
* `File.exists?` is deprecated

This does not address the "possible reference to past scope" warnings
introduced with Ruby 2.2.0 as it appears this warning will be removed in
2.2.1. See https://bugs.ruby-lang.org/issues/10661

### DIFF
--- a/spec/models/password_strategies_spec.rb
+++ b/spec/models/password_strategies_spec.rb
@@ -3,6 +3,8 @@ require 'spec_helper'
 describe Clearance::User do
   subject do
     class UniquenessValidator < ActiveModel::Validator
+      undef validate
+
       def validate(record)
       end
     end
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -27,7 +27,7 @@ module Clearance
     end
 
     def user_model
-      @user_model || ::User
+      @user_model ||= ::User
     end
 
     def allow_sign_up?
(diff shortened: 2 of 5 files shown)


spec/models/password_strategies_spec.rb
```diff
@@ -3,6 +3,8 @@
 describe Clearance::User do
   subject do
     class UniquenessValidator < ActiveModel::Validator
+      undef validate
```
jferris: Does Ruby warn about overriding a method from a superclass?

jferris:
Looks good to merge.

spec/models/password_strategies_spec.rb
derekprior: I couldnt really figure this out. I think it may have to do with this being a "subject" which gets continually re-evaluated for each spec.

derekprior:
This doesn't actually appear to be all the warnings as there seems to be a difference between `rspec --warnings` (a lot more warnings) and `rspec` with `config.warnings = true`. It seems the latter just gives you the warnings once the suite starts, perhaps?

derekprior:
Regardless, merged.

spec/models/password_strategies_spec.rb
jferris: Oh, I missed that; that's definitely what it is. You're defining this class every time the test runs. You could use `stub_const` from RSpec, which will then clean it up after the test.

Issue #503, opened 2015-01-05
Compatibility with Pundit?


andrewhavens:
I am using [Pundit](https://github.com/elabs/pundit) for authorization in my application. I just added Clearance and changed my controller `before_action` to use `:authorize`. However, I now realize that Pundit uses a controller method by the same name. So Pundit's `authorize` method is being called instead of the Clearance `authorize` method...what can I do to get around this issue?

andrewhavens:
Ah, sorry, I should have searched first. This question was answered in issue #436. It sounds like the plan is to rename this method in the next major version of Clearance. In the mean time, you can alias the method to something else in your `ApplicationController`:

```ruby
class ApplicationController < ActionController::Base
  include Clearance::Controller

  alias_method :require_login, :authorize

  include Pundit
end

class PostsController < ApplicationController
  before_filter :require_login, only: [:edit, :update]

  def edit
    @post = Post.find(params[:id])
    authorize @post
  end
end
```

Issue #436, opened 2014-06-04
Clearance not compatible with Pundit


jessieay:
Clearance and [Pundit](https://github.com/elabs/pundit) both use `authorize`. I was just implementing Pundit on a project with Clearance auth, and the solution for getting Pundit and Clearance to both work was to recreate `authorize` in our `ApplicationController` and name it `clearance_authorize`.

While renaming the method was easy, this also required us to override all Clearance controllers since the `skip_before_authorize` actions were still skipping `authorize` rather than `clearance_authorize`. 

A way to fix Clearance's incompatibility with Pundit would be to rename `authorize` so that it does not clash. Also open to other ideas! (I will write the code / open a PR if a general solution is agreed upon beforehand)

Issue #239, opened 2012-12-13
Authenticate vs Authorize


halogenandtoast:
Authentication and Authorization are two completely different things. I'm continuously  frustrated and confused by this each time I used clearance. I'll submit a pull request later deprecating the old way and adding in the new suggested way.

PR #506, opened 2015-01-06
Rename `authorize` filter to `require_login`

derekprior:
This name better expresses the intent of the filter and has the advantage of
not conflicting with the `authorize` method provided by pundit or wading into
the sometimes hairy line demarcating authorization from authentication.

Clearance users should migrate from `:authorize` to `require_login` as the
former will be removed in 2.0. Be sure to catch reference to
`skip_before_filter :authorize` or `skip_before_action :authorize`, which the
deprecation cannot catch.

addresses #503, #436, and #239

### DIFF
--- a/lib/clearance/authorization.rb
+++ b/lib/clearance/authorization.rb
@@ -3,13 +3,21 @@ module Clearance
     extend ActiveSupport::Concern
 
     included do
-      hide_action :authorize, :deny_access
+      hide_action :authorize, :deny_access, :require_login
     end
 
-    def authorize
+    def require_login
       unless signed_in?
         deny_access
       end
+    end
+
+    def authorize
+      warn "[DEPRECATION] Clearance's `authorize` before_filter is " +
+        "deprecated. Use `require_login` instead. Be sure to update any " +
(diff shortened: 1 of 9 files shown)


andrewhavens:
:+1:

arthurnn:
LGTM

gabebw:
Love it. The `authorize` name always kind of bugged me, since Clearance is an authentication library, not an authorization library

lib/clearance/authorization.rb
```diff
@@ -3,15 +3,21 @@ module Authorization
...
 
-    def authorize
+    def require_login
       unless signed_in?
         deny_access
       end
     end
 
+    def authorize
+      warn "[DEPRECATION] Clearance's `authorize` before_filter is deprecated."\
```
gabebw: Why not use `+` instead of `\` so it's just basic string concatenation?

derekprior: Updated the deprecation to use concatenation and be more explicit about the potential gotcha with `skip_before_filter`

PR #507, opened 2015-01-07
Deprecate `Clearance.root`

derekprior:
This appears to have been a convenience method used only in a single
generator. While the replacement is uglier it has the distinct advantage of
not leaking into the public API of the gem.

### DIFF
--- a/lib/clearance.rb
+++ b/lib/clearance.rb
@@ -11,6 +11,9 @@ require 'clearance/constraints'
 
 module Clearance
   def self.root
+    warn "#{Kernel.caller.first}: [DEPRECATION] `Clearance.root` is " +
+      "deprecated and will be removed in the next major release. If you need " +
+      "to find Clearance's root, you can use the `Gem::Specification` API."
     File.expand_path('../..', __FILE__)
   end
 end
--- a/lib/generators/clearance/views/views_generator.rb
+++ b/lib/generators/clearance/views/views_generator.rb
@@ -3,7 +3,7 @@ require 'rails/generators/base'
 module Clearance
   module Generators
     class ViewsGenerator < Rails::Generators::Base
-      source_root Clearance.root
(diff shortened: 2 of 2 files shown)


mike-burns:
Does this change the major version?

mike-burns:
Ah, you're bumping it anyway. Carry on.

derekprior:
I don't think it should bump the major version. I can't think of a reason why anyone would be using this this method. It doesn't appear anywhere in the documentation though, to be fair, this Gem has no documentation other than a README. That's actually how I came across it, I opened it up to start adding RDoc.

I could deprecate it instead, which would be safer, but I just don't think it matters. Thoughts?

derekprior:
> Ah, you're bumping it anyway. Carry on.

Eventually. There are still many 1.x releases until 2.0. I plan to deprecate things I consider to be "actual for real public API" (basically anything other than this method here).

lib/clearance.rb
```diff
@@ -10,7 +10,4 @@
 require 'clearance/constraints'
 
 module Clearance
```
gabebw: Can we remove this `module` too, since everything is already loaded by this point?

derekprior: I was thinking it'd be a good place to add any top-level RDoc, but maybe I'm wrong. I've never RDoc'd anything.

gabebw: Oh, probably. Then I'm evenly split between "Just remove this and add it when you start adding RDoc" and "Might as well leave it in."

mike-burns:
My thoughts on bumping the major: we have complained every time the minor was bumped but the API changed. Since this method was as documented as many actual API methods, it's part of the API. Therefore, someone will complain.

I'd like to live in a society where removing a public method causes a major bump.

derekprior:
That's fair. I'll deprecate.

PR #509, opened 2015-01-08
Test & Support Ruby 2.2 across Rails versions

derekprior:
To get the test suite to run in under 2.2, I:

* Upgraded Cucumber to a version that supports Ruby 2.2
* Added 2.2.0 to our Travis Matrix (Sorry, Travis)
* Excluded Rails 3.2 under Ruby 2.2 from appraisals as it is not
  supported by rails (yet). See: rails/rails#18306

Once that was done, it was  discovered that Rails 4.0.x requires the
`test-unit` gem under Ruby 2.2. Adding that gem allows the test suite to
run there.

With that in place, I found that the `deny_access` matcher was not
negating as expected. This is because the test-unit gem raises a
different error when an assertion failed. I have to catch this error in
addition to the Minitest::Assertion error we were already catching.


gabebw:
:+1:

iwz:
:+1:

Issue #453, opened 2014-08-18
Sign in page can be visited while signed in


gabrielecirulli:
Using the latest gem version of clearance, without any strange configuration and tweaks (except I generated the views and tweaked them), I can visit the sign in page while I'm already signed in.

The expected behavior would be being redirected to either the root path or a path of my liking if I try to visit that page when signed in.

Is this possible without having to override things?

derekprior:
There's an argument to be made that this is just a bug and should be fixed. I can't think of any use case that this would totally destroy, so I think we should probably just fix it.

gabrielecirulli:
I agree. The general consensus on the web is to "bounce" back users that are already signed in if they try to visit a page that is reserved for signed-out users. This issue is not necessarily a bug (it doesn't cause any problems), but it can definitely cause confusion and make an application not feel very solid to the user.

derekprior:
Closed with #513, which will be in our next 1.x release (likely 1.7.1, possibly 1.8).

PR #513, opened 2015-01-09
Redirects user to redirect_url if signed in

kenyonj:
https://github.com/thoughtbot/clearance/issues/453

### DIFF
--- a/app/controllers/clearance/sessions_controller.rb
+++ b/app/controllers/clearance/sessions_controller.rb
@@ -1,4 +1,5 @@
 class Clearance::SessionsController < Clearance::BaseController
+  before_filter :redirect_signed_in_users, only: [:new]
   skip_before_filter :require_login, only: [:create, :new, :destroy]
   protect_from_forgery except: :create
 
@@ -10,7 +11,7 @@ class Clearance::SessionsController < Clearance::BaseController
         redirect_back_or url_after_create
       else
         flash.now.notice = status.failure_message
-        render template: 'sessions/new', status: :unauthorized
+        render template: "sessions/new", status: :unauthorized
       end
     end
   end
@@ -21,10 +22,16 @@ class Clearance::SessionsController < Clearance::BaseController
   end
 
   def new
-    render template: 'sessions/new'
+    render template: "sessions/new"
   end
 
   private
+
+  def redirect_signed_in_users
+    if signed_in?
+      redirect_to url_for_signed_in_users
+    end
+  end
 
   def url_after_create
     Clearance.configuration.redirect_url
@@ -32,5 +39,9 @@ class Clearance::SessionsController < Clearance::BaseController
 
   def url_after_destroy
     sign_in_url
+  end
+
+  def url_for_signed_in_users
+    url_after_create
   end
 end
--- a/spec/controllers/sessions_controller_spec.rb
+++ b/spec/controllers/sessions_controller_spec.rb
@@ -1,19 +1,31 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe Clearance::SessionsController do
   it { should be_a Clearance::BaseController }
 
-  describe 'on GET to /sessions/new' do
-    before { get :new }
+  describe "on GET to #new" do
+    context "when a user is not signed in" do
+      before { get :new }
 
-    it { should respond_with(:success) }
-    it { should render_template(:new) }
-    it { should_not set_the_flash }
+      it { should respond_with(:success) }
+      it { should render_template(:new) }
+      it { should_not set_the_flash }
+    end
+
+    context "when a user is signed in" do
+      before do
+        sign_in
+        get :new
+      end
+
+      it { should redirect_to(Clearance.configuration.redirect_url) }
+      it { should_not set_the_flash }
+    end
   end
 
-  context 'when password is optional' do
-    describe 'POST create' do
-      it 'renders the page with error' do
+  describe "on POST to #create" do
+    context "when password is optional" do
+      it "renders the page with error" do
         user = create(:user_with_optional_password)
 
         post :create, session: { email: user.email, password: user.password }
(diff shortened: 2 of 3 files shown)


app/controllers/clearance/sessions_controller.rb
```diff
@@ -21,7 +21,11 @@ def destroy
   end
 
   def new
-    render template: 'sessions/new'
+    if current_user
```
derekprior: Can we do this as a before_filter instead? That will make it much simpler to override as you'll just be able to `skip_before_filter :redirect_signed_in_users` or whatever its called.

app/controllers/clearance/sessions_controller.rb
```diff
@@ -33,4 +37,8 @@ def url_after_create
   def url_after_destroy
     sign_in_url
   end
+
+  def redirect_url
```
derekprior: If this is going to be a separate method, it should probably be named consistently with the other controller methods available for override [here](https://github.com/thoughtbot/clearance#overriding-controllers).

But I think maybe we should just use `url_after_create` just as if their sign in had already been successful?

spec/controllers/sessions_controller_spec.rb
```diff
@@ -4,11 +4,23 @@
   it { should be_a Clearance::BaseController }
 
   describe 'on GET to /sessions/new' do
-    before { get :new }
+    context 'when a user is not signed in' do
```
derekprior: I'm surprised hound hasn't complained, but we've been using double quotes now.

app/controllers/clearance/sessions_controller.rb
kenyonj: ok, I was not sure if another method was needed since it was somewhat incorrect about what the url was actually for

spec/controllers/sessions_controller_spec.rb
kenyonj: oh it has been... but the whole file was already in single quotes so I thought that is what the project was using... I will change all to double now.

app/controllers/clearance/sessions_controller.rb
derekprior: I'd also be cool with "url_for_users_already_signed_in" which defaults to `url_after_create`? Something like that. That gives maximum flexibility.

derekprior: `url_for_signed_in_users`

spec/controllers/sessions_controller_spec.rb
```diff
@@ -22,64 +34,66 @@
...
+      it "should reset the remember token" do
+        expect(@user.reload.remember_token).not_to eq "old-token"
+      end
+
+      it "should unset the current user" do
+        expect(@controller.current_user).to be_nil
+      end
     end
   end
 end
```
kenyonj: @derekprior I have reorganized the spec file. It seemed to be using contexts in some places and not others.

app/controllers/clearance/sessions_controller.rb
```diff
@@ -33,4 +40,8 @@ def url_after_create
   def url_after_destroy
     sign_in_url
   end
+
+  def url_for_signed_in_users
+    url_after_create
```
andrewhavens: Would it be possible to make this a configuration value like `url_after_create` does? Now that there are multiple types of redirects, seems like `Clearance.configuration.redirect_url` should be renamed to something like `Clearance.configuration.after_create_redirect_url` and add a `Clearance.configuration.after_sign_in_redirect_url`

spec/controllers/sessions_controller_spec.rb
```diff
@@ -1,19 +1,31 @@
...
+      it { should_not set_the_flash }
+    end
+
+    context "when a user is signed in" do
+      before do
+        sign_in
+        get :new
+      end
+
+      it { should redirect_to(Clearance.configuration.redirect_url) }
```
andrewhavens: This is the line that made me think about the configuration values. When a user is signed in, it should not go to the `url_after_create` but the `url_for_signed_in_users`. They happen to be the same value in this case since the defaults are the same. But the expectation should read that it redirects to the "already signed in" URL.

app/controllers/clearance/sessions_controller.rb
derekprior: I'm hesitant to add a configuration option dedicated for this purpose. It's already factored out into a method which can be overridden if desired and is ultimately backed by `Clearance.configuration.redirect_url`.

Is there a use case where you would expect these two URLs to be different?

derekprior:
Thanks @kenyonj. This looks good to me but it looks like it should remain open for some discussion for a bit.

app/controllers/clearance/sessions_controller.rb
kenyonj: There are several methods that point to `Clearance.configuration.redirect_url` and in this instance I think we actually want to redirect to the same place that users get redirected to after the session is created. This can always be overridden in the controller.

andrewhavens: `url_for_signed_in_users` makes sense to redirect to the `Clearance.configuration.redirect_url`. I think that is the most common configuration behavior. However, I think the reason for this PR is to be able to have two different redirect URLs depending on if you are a newly created user, or a user who has already logged in. Oh wait...maybe I'm reading this wrong. My thought was that I want to be able to specify a different redirect URL for a newly created user (taking them to a "fill out additional profile information" page). But maybe `create` in this context means the creation of a new session.

andrewhavens: Looks like there is already a [`UsersController#url_after_create`](https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/users_controller.rb#L27). My mistake! Sorry about that.

kenyonj: no problem. yes, this is for signing in, so in this context `create` just means that the user is signing in.

derekprior: I think it's possible we may want to clarify `Clearance.configuration.redirect_url` in the future, but I don't think it's something we have to do right now.

>  I think the reason for this PR is to be able to have two different redirect URLs depending on if you are a newly created user, or a user who has already logged in. Oh wait...maybe I'm reading this wrong. My thought was that I want to be able to specify a different redirect URL for a newly created user (taking them to a "fill out additional profile information" page). But maybe create in this context means the creation of a new session (i.e. after sign in v.s. visiting sign in page then automatically redirecting if you're signed in).

This is the sessions controller, not the users controller. The intent of this PR is to prevent already signed in users from reaching the log in page, which might be confusing. Though that raises an interesting question -- do we have this problem on the new user page as well?

spec/controllers/sessions_controller_spec.rb
kenyonj: We are asserting that the user should be redirected to the `redirect_url` regardless of what the methods in the controller are set to. This is the default and we want this test to fail if the code is changed that implements this fix.

app/controllers/clearance/sessions_controller.rb
```diff
@@ -21,11 +22,17 @@ def destroy
...
 
   def new
-    render template: 'sessions/new'
+    render template: "sessions/new"
   end
 
   private
 
+  def redirect_signed_in_users
+    if current_user
```
derekprior: `signed_in?` is probably better here. I think this will break if you have a guest user, for instance.

derekprior:
merge away!

PR #514, opened 2015-01-09
Replace Cucumber with RSpec acceptance specs

derekprior:
The replacement spec does much the same as the features were doing before, but
in a (hopefully) more maintainable manner. We generate a fresh rails app,
install clearance, and run clearance's generated specs.

I kept the acceptance spec in its own suite because the output of it is a bit
odd and I wanted it separate from our usual tests. It will output the results
of the generated specs followed by the result of the single acceptance spec.


spec/acceptance/clearance_installation_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "spec_helper"
+
+describe "Clearance Installation" do
+  around do |example|
+    current_path = File.expand_path("../", __FILE__)
+    Dir.chdir("tmp")
```
sikachu: If example raise an exception, there's a chance that line 11 won't get run, right? How about using the block version of `chdir`?

sikachu: Not sure if it matters much though, but I assume that your CWD will leak to the next run.

derekprior: I didn't know about or had forgotten the block version. Thanks.

PR #515, opened 2015-01-09
Deprecates avoid_sign_in

kenyonj:
This also refactors the `users_controller_spec` to follow the `expect` syntax.

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,88 +1,75 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe Clearance::UsersController do
   it { should be_a Clearance::BaseController }
 
-  describe 'when signed out' do
-    before { sign_out }
+  describe "on GET to #new" do
+    context "when signed out" do
+      it "renders a form for a new user" do
+        get :new
 
-    describe 'on GET to #new' do
-      before { get :new }
-
-      it { is_expected.to respond_with(:success) }
-      it { is_expected.to render_template(:new) }
-      it { is_expected.not_to set_the_flash }
-    end
-
-    describe 'on GET to #new with email' do
-      before do
-        @email = 'a@example.com'
-        get :new, user: { email: @email }
+        expect(response).to be_success
+        expect(response).to render_template(:new)
       end
 
-      it 'should set assigned user email' do
-        expect(assigns(:user).email).to eq @email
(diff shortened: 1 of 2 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -1,88 +1,84 @@
...
+require "spec_helper"
 
 describe Clearance::UsersController do
   it { should be_a Clearance::BaseController }
 
-  describe 'when signed out' do
-    before { sign_out }
+  describe "on GET to #new" do
+    context "when signed out" do
+      context "with email" do
```
derekprior: I'd get rid of this context and change the order of these two it blocks to say something like:

```
it "renders a form for a new user"
it "renders a form for a user with the query-string-provided email"
```

As a general rule don't use "should" in the it block description.

derekprior: Maybe a better description of the email one is:

```
it "defaults email to value provided in query string"
```

spec/controllers/users_controller_spec.rb
```diff
@@ -1,88 +1,84 @@
...
-        @old_user_count = User.count
-        @return_url = '/url_in_the_session'
-        @request.session[:return_to] = @return_url
-        post :create, user: user_attributes
-      end
+  describe "on POST to #create" do
+    context "when signed out" do
+      context "with valid attributes" do
+        it "assigns and creates a user then redirects to the redirect_url" do
+          sign_out
```
derekprior: are the `sign_out` calls necessary?

spec/controllers/users_controller_spec.rb
```diff
@@ -1,88 +1,84 @@
...
+          expect(User.count).to eq old_user_count + 1
+          expect(response).to redirect_to(return_url)
+        end
       end
     end
 
-    describe 'POST to create' do
-      before { post :create, user: {} }
+    context "when signed in" do
+      it "redirects to the home page" do
```
derekprior: redirects to the configured clearance redirect url.

derekprior:
Few minor comments. Thanks for updating the spec as well.

spec/controllers/users_controller_spec.rb
kenyonj: no they are not... they were brought over from the previous `before` blocks... I will remove them

derekprior:
LGTM. Assuming travis passes, squash (2 commits looks good here) and merge.

PR #516, opened 2015-01-09
Use dummy app for rails dependent specs

jessieay:
* `User` was previously defined in clearance.rb, which was confusing
* Move app_templates out of spec/support directory to avoid `User` class being
 `require`d twice


spec/generators/clearance/install/install_generator_spec.rb
```diff
@@ -52,7 +52,6 @@
         expect(user_class).to exist
         expect(user_class).to have_correct_syntax
         expect(user_class).to contain("include Clearance::User")
-        expect(user_class).to have_method("previously_existed?")
```
derekprior: This was testing that the Clearance::User module was included to a previously-existing user-class (rather than the generator creating a new class). I don't think this tests the same thing without this. Granted, the assertion here is rather magical. Not sure what else to do though.

derekprior:
Looks good. Rebase and merge.

Issue #517, opened 2015-01-14
:authorize rendered or redirected


gabrieltong:
Hey 
I updated my gem today , clearance update from 1.4 to 1.7 , but I face this redirect problem , can you help me ? 

thanks man !

```
Started GET "/sign_in" for 127.0.0.1 at 2015-01-14 15:35:15 +0800
Processing by Clearance::SessionsController#new as HTML
Redirected to http://localhost:4000/sign_in
Filter chain halted as :authorize rendered or redirected
Completed 302 Found in 1ms (ActiveRecord: 0.0ms)
[DEPRECATION] Clearance's `authorize` before_filter is deprecated. Use `require_login` instead. Be sure to update any instances of `skip_before_filter :authorize` or `skip_before_action :authorize` as well


Started GET "/sign_in" for 127.0.0.1 at 2015-01-14 15:35:15 +0800
Processing by Clearance::SessionsController#new as HTML
Redirected to http://localhost:4000/sign_in
Filter chain halted as :authorize rendered or redirected
Completed 302 Found in 1ms (ActiveRecord: 0.0ms)
[DEPRECATION] Clearance's `authorize` before_filter is deprecated. Use `require_login` instead. Be sure to update any instances of `skip_before_filter :authorize` or `skip_before_action :authorize` as well


Started GET "/sign_in" for 127.0.0.1 at 2015-01-14 15:35:15 +0800
Processing by Clearance::SessionsController#new as HTML
Redirected to http://localhost:4000/sign_in
Filter chain halted as :authorize rendered or redirected
Completed 302 Found in 1ms (ActiveRecord: 0.0ms)
[DEPRECATION] Clearance's `authorize` before_filter is deprecated. Use `require_login` instead. Be sure to update any instances of `skip_before_filter :authorize` or `skip_before_action :authorize` as well


Started GET "/sign_in" for 127.0.0.1 at 2015-01-14 15:35:15 +0800
Processing by Clearance::SessionsController#new as HTML
Redirected to http://localhost:4000/sign_in
Filter chain halted as :authorize rendered or redirected
Completed 302 Found in 1ms (ActiveRecord: 0.0ms)
```

derekprior:
Have you overridden the sessions controller at all? If so, can you share it
here?

gabrieltong:
no, I doesn't change session controller 

but it's fixed by replacing `authorize` by `require_login` .

derekprior:
[...]

PR #518, opened 2015-01-14
Prevent redirect loops when still using `authorize`

derekprior:
The `authorize` before filter is deprecated in favor of `require_login`.
However, just switching Clearance's internal filters is not sufficient. We're
dependent on users updating any `before_filter :authorize` calls.

If they still have `before_filter :authorize` in their application controller,
then they will see a deprecation on calls to authorize but the method will
still run (it's aliased to `require_login`). This may cause them to be
redirected to sign in. Sign in is set to
`skip_before_filter :require_login`, but will happily still run authorize if
instructed to by the application controller. Then you're stuck in a redirect
loop.

By duplicating the `skip_before_filter` calls to also skip `authorize` we can
be sure this doesn't happen.

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -2,6 +2,7 @@ require 'active_support/deprecation'
 
 class Clearance::PasswordsController < Clearance::BaseController
   skip_before_filter :require_login, only: [:create, :edit, :new, :update]
+  skip_before_filter :authorize, only: [:create, :edit, :new, :update]
   before_filter :forbid_missing_token, only: [:edit, :update]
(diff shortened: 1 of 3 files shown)


derekprior:
See also #517

gabebw:
Can we write tests for this, or this is inherently untestable?

derekprior:
I think testing it would be difficult, but not impossible. I'd need to temporarily create an application controller (since that's what the clearance controllers inherit from) that uses the `authorize` filter then be sure I don't get stuck in a redirect loop.  

Not sure of the best way to do that because I can't easily undo the side effects of `before_filter`. That is, I can't simply undef a method when I'm done.

gabebw:
I suppose this is OK then.

Issue #465, opened 2014-09-22
Expire password reset token after a period of time


aripollak:
Generating a password reset token without any expiration time opens up a potential security flaw. Not a very easily exploitable one, but an easily preventable one nonetheless. More info: http://www.owasp.org/index.php/Forgot_Password_Cheat_Sheet#Step_3.29_Send_a_Token_Over_a_Side-Channel
In addition to their email account being compromised, there's another possible scenario: if a user clicks Reset Password but never gets around to actually resetting their password, anyone who gets access to the backend database at any point in the future can reset that user's password. This would be much faster than trying to crack the encrypted password. And since by default there is no email sent upon password change, the user might not ever know that their account has been compromised.

nsuchy:
Is this still an issue?

PR #521, opened 2015-01-16
Do not validate user when resetting password

### LINKED ISSUE #406: Don't validate user when updating password
When updating a users password, clearance [runs a full validation](https://github.com/thoughtbot/clearance/blob/master/lib/clearance/user.rb#L78). There are a number of reasons that a user may be invalid that are out of the scope of what clearance provides. Password resets should not be held up due to a new required field being added to user, for instance.

derekprior:
User records can be invalid for many reasons that are not central to
Clearance's concerns. An application may decide at some point to begin
requiring all user records to have a zip code. They reasonably assume this
will effect only new registrations or existing users who attempt to edit
their profile. Unfortunately, it would have also prevented users from
resetting their password because we were running validations as part of
that operation. With zip code not present on the password reset form, the
user would have no way to complete the password reset.

This change still calls `valid?` but considers only errors that are on the
`password` field when deciding whether to proceed. This allows for
additional validations on `password`, such as length. Ultimately, the save
is done without validations assuming `password` itself is valid.

The nested `before` blocks in the specs caused a great deal of frustration when
adding tests for this functionality. The specs were re-written in a style
consistent with how we write specs today, taking care to ensure tests still
fail appropriately.

Closes #406


bin/rspec
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+#
+# This file was generated by Bundler.
```
MattMSumner: I know this is generated, is it worth removing comments?

lib/clearance/user.rb
```diff
@@ -75,13 +75,13 @@ def reset_remember_token!
     def update_password(new_password)
       self.password_changing = true
       self.password = new_password
+      valid?
 
-      if valid?
+      if errors[:password].blank?
```
MattMSumner: could you use `if valid? || errors[:password].blank?`?

MattMSumner:
couple of comments otherwise LGTM

lib/clearance/user.rb
derekprior: After a bit of thought, yes. I like that it's one line but I'm not sure it's more clear. I had to think through the truth table a bit.

It is odd to have the query method `valid?` being used as a command though. Hmm...

bin/rspec
derekprior: I'm happy neough to leave them in just in case someone unfamiliar comes across them and is confused or decides to edit them manually.

lib/clearance/user.rb
calebthompson: `validate` is available in newer versions of Rails. Just a bit of trivia that isn't useful to you here.

calebthompson: It's less a command than you're letting on, though. If this is `valid?`, there's no reason to check that `errors[:password].blank?` (would `empty?` make more sense here?)

MattMSumner: that's more what I was thinking. If `valid?` is true there's no need to check errors.

mxie: AFAIK, the purpose of calling `valid?` here is to run the validations, not so much to actually use its return value. Without `valid?`, `errors` wouldn't be populated.

mxie: I think we care more about `valid?` as a command here than its actual return value, so I'm going to side with Derek on keeping these separate. I do, however, think that it makes more sense to use `empty?` instead of `blank?`.

derekprior: I'm not concerned with save an in-memory hash lookup when resetting a password. I just want it to be as clear as possible. 

with regards to `blank?` or `empty?`, `blank?` seems safer. If for whatever reason that's `nil`, `blank?` still works as expected while `empty?` does not.

derekprior: Some quick testing shows that `empty?` really should be safe though. Even looking for non-existent attributes still returns an empty array.

mxie: It looks like for ActiveModel, `empty?` is just aliased as `blank?`, so it makes no difference.

lib/clearance/user.rb
```diff
@@ -75,13 +75,13 @@ def reset_remember_token!
...
         self.confirmation_token = nil
         generate_remember_token
+        save(validate: false)
```
jferris: Should you make sure to only update the relevant password attributes here, since we're not sure the others are valid?

derekprior: They all need to be updated. To complete the password rotation. I can't
split it into 2 calls to save (one with validate false) in a transaction
because the second save would be subject to the same validation issues we
have now.

Best I can do is to set them before calling valid and checking for errors
on those fields too.

derekprior: Oh I see what you're saying - be explicit about the columns I'm updating so as not to allow something like:

```
user.email = ""
user.update_password("foo")
```

jferris: Right. The only way I know of to do this is with `update_columns`.

This would be a breaking change, but since it seems like the save/validate lifecycle of users/passwords is different, one thing you could do is split out a new table/model to represent authentication tokens and make them belong to a user. That way, you could expire/delete old tokens and create new ones without touching the user/model table and its maze of validations.

calebthompson: I'm in support of anything getting engine-specific stuff out of my User.

If we don't go that route, I just wanted to note that `update_columns` won't touch `update_at`, which is a good or bad thing depending on how you look at it.

derekprior: Can't I call `user.update_attributes(password: new_password, confirmation_token: user.confirmation_token, remember_token: user.remember_token)`. It's odd looking but I think it would work, right?

> This would be a breaking change, but since it seems like the save/validate lifecycle of users/passwords is different, one thing you could do is split out a new table/model to represent authentication tokens and make them belong to a user. That way, you could expire/delete old tokens and create new ones without touching the user/model table and its maze of validations.

This is already in the plans for addressing #465. The plan for that is:

1. Extract password reset to a service and make the methods in `Clearance::User` passthroughs to that service, deprecate direct calls.
2. Expose a new service that splits out password resets, make this the default for newly installed apps in the initializer.
3. Deprecate the old service.

calebthompson: Naw, `update_attributes` calls `assign_attributes` and `save`, so any pre-existing assignments would have also been committed.

derekprior: Oh, no way to add `validate: false` to `update_attributes`.

jferris: `update_columns` will work the same way, but without touching other attributes (including timestamps) or going through validations or callbacks.

If you want a way to update a table without messing with the model, `update_columns` is what you're looking for.

derekprior: Ok, how about using `changes` to verify only the columns we want are updated. If there are more columns we fall back to `save`, otherwise `save validate: false`?

derekprior: > update_columns will work the same way, but without touching other attributes (including timestamps) or going through validations or callbacks.

I'm not sure I can safely assume they don't want callbacks run or timestamps updated.

derekprior:
I'm going to merge in the spec improvements but leave the functionality unchanged. @mxie is working on a new password reset mechanism and I think given the problems in implementing this here it's reasonable for this issue to be addressed only there.

PR #522, opened 2015-01-22
Update Passwords Controller Spec

derekprior:
The nested `before` blocks caused a great deal of frustration when adding
additional tests to password reset functionality. The specs were re-written in
a style consistent with how we write specs today, taking care to ensure tests
still fail appropriately.


jferris:
These look way better to me, and nothing looks wrong.

I think some of the `it` examples within `context` groups could be safely merged without losing any readability.

mxie:
These look good to me too.

PR #523, opened 2015-01-22
Configure remember token cookie name

tabfugnic:
Gives the user full configurability when dealing with apps with similar paths.

### DIFF
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -2,8 +2,6 @@ require 'clearance/default_sign_in_guard'
 
 module Clearance
   class Session
-    REMEMBER_TOKEN_COOKIE = 'remember_token'.freeze
-
     def initialize(env)
       @env = env
       @current_user = nil
@@ -12,7 +10,11 @@ module Clearance
 
     def add_cookie_to_headers(headers)
       if cookie_value[:value].present?
-        Rack::Utils.set_cookie_header!(headers, REMEMBER_TOKEN_COOKIE, cookie_value)
+        Rack::Utils.set_cookie_header!(
+          headers,
+          remember_token_cookie,
+          cookie_value
+        )
       end
     end
 
@@ -29,7 +31,7 @@ module Clearance
       status = run_sign_in_stack
 
       if status.success?
-        cookies[REMEMBER_TOKEN_COOKIE] = user && user.remember_token
+        cookies[remember_token_cookie] = user && user.remember_token
       else
         @current_user = nil
       end
@@ -45,7 +47,7 @@ module Clearance
       end
 
       @current_user = nil
-      cookies.delete REMEMBER_TOKEN_COOKIE
+      cookies.delete remember_token_cookie
     end
 
     def signed_in?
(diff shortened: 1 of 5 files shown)


spec/clearance/session_spec.rb
```diff
@@ -30,6 +30,17 @@
     expect(session.current_user).to be_nil
   end
 
+  it 'sets a custom cookie name' do
+    Clearance.configuration.cookie_domain = 'custom_token'
```
tabfugnic: I noticed this project uses primarily single quotes, are you slowly moving this over to double quotes or should I leave this?

derekprior: moving over.

spec/clearance/session_spec.rb
```diff
@@ -30,6 +30,17 @@
...
+
+    session.sign_in user
+    session.add_cookie_to_headers(headers)
+
+    expect(headers['Set-Cookie']).to match(/custom_token/)
+
+    restore_default_config
```
derekprior: This won't execute if the expectation fails. I think for now it's fine to keep with the existing specs and do this with an `after`.

derekprior:
There's a section of the readme that lists each configuration option. Can you add this there?

lib/clearance/session.rb
```diff
@@ -2,7 +2,7 @@
 
 module Clearance
   class Session
-    REMEMBER_TOKEN_COOKIE = 'remember_token'.freeze
+    REMEMBER_TOKEN_COOKIE = Clearance.configuration.cookie_name.freeze
```
jferris: I think setting this here as a constant will mean that it won't pick up the change if you change it in something like an initializer, because the Clearance middleware stack will be eagerly loaded. Is that right?

tabfugnic: That's a good question, should this then be thrown into the initializer?

jferris: Can we remove this constant entirely, or is it considered part of the public API?

derekprior: I don't have a problem removing it. I wouldn't consider it a breaking change.

lib/clearance/session.rb
```diff
@@ -62,8 +64,12 @@ def cookies
       @cookies ||= @env['action_dispatch.cookies'] || Rack::Request.new(@env).cookies
     end
 
+    def remember_token_cookie
```
jferris: There are a few other methods later which are related to looking at the Clearance config. What do you think about moving this method down there?

jferris:
I had one minor comment, but I think this is good to merge whenever you're ready.

@derekprior any thoughts?

derekprior:
Looks good to me.

tabfugnic:
@derekprior All taken care of. What's procedure for getting this merged in?

derekprior:
Squash and merge (ff-only).

I can release this as 1.8, though I have to write up the NEWS for a few changes that will be included. I'll probably do that tonight. If there's a pressing need for it to be released now, let me know.

tabfugnic:
Not hugely important, but sooner rather than later would be great

derekprior:
1.8.0 is released.

PR #524, opened 2015-01-22
Decouple tests for resetting and updating passwords

mxie:
These methods should be tested separately instead of nested within each other.


derekprior:
:+1:

PR #525, opened 2015-01-23
v1.8.0

derekprior:


NEWS.md
```diff
@@ -1,5 +1,16 @@
 Thank you to all the [contributors](https://github.com/thoughtbot/clearance/graphs/contributors)!
 
+New for 1.8.0 (January 23, 2015)
+* Fixed an issue that would cause sites still using the deprecated `authorize`
```
sikachu: "site **that** still using"?

Feel free to leave it if that's already grammatically correct.

NEWS.md
```diff
@@ -1,5 +1,16 @@
...
+  filter to enter a redirect loop when redirecting to the sign in path.
+* The Clearance remember token cookie name is now customizable via
+  `Clearance.configuration.cookie_name`.
+* Signed in users that attempt to visit the sign in path are now redirected. The
+  redirect URL defaults to the same URL used for the redirect after sign in, but
+  can be customized by overriding `passwords_controller#url_for_signed_in_users`
+* `users_controller#avoid_sign_in` is now deprecated in favor of
```
sikachu: What happen if they define `#avoid_sign_in`? Does it still work?

derekprior: `avoid_sign_in` was never documented as part of the public API. The deprecation is out of abundance of caution. If they override that method, their override will not be called (unless they add the filter), but `redirect_signed_in_users` does something logical and non-dangerous anyway.

derekprior: Basically, I expect no one to get this deprecation.

sikachu: I see. So that's not a breaking change. :+1:

PR #526, opened 2015-01-30
Add documentation for configuration options

derekprior:
Added Yard configuration and documentation for
`Clearance::Configuration`. Moving forward, we will be adding
documentation to the rest of Clearance.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,0 +1,3 @@
+--exclude templates
+--markup markdown
+--markup-provider redcarpet
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -1,23 +1,94 @@
 module Clearance
   class Configuration
-    attr_writer :allow_sign_up, :routes
+    # Controls whether the sign up route is enabled.
+    # Defaults to `true`. Set to `false` to disable user creation routes.
+    # The setting is ignored if routes are disabled.
+    # @param [Boolean] value
+    # @return [Boolean]
+    attr_writer :allow_sign_up
 
-    attr_accessor \
-      :cookie_domain,
-      :cookie_expiration,
-      :cookie_name,
-      :cookie_path,
-      :httponly,
-      :mailer_sender,
-      :password_strategy,
-      :redirect_url,
-      :secure_cookie,
-      :sign_in_guards,
-      :user_model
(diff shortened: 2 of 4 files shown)


lib/clearance/configuration.rb
```diff
@@ -44,23 +121,36 @@ def  user_actions
...
   end
 
+  # Set Clearance's configuration
+  # @param config [Clearance::Configuration]
   def self.configuration=(config)
     @configuration = config
   end
 
+  # Modify Clearance's current configuration
+  #   Clearance.configure do |config|
```
mcmire: This code sample might not render correctly, you might have to indent this by four spaces to do so.

lib/clearance/configuration.rb
```diff
@@ -1,23 +1,94 @@
...
+    # live environments to prevent session hijacking. For more, see
+    # [RFC6265](http://tools.ietf.org/html/rfc6265#section-5.2.5).
+    # @return [Boolean]
+    attr_accessor :secure_cookie
+
+    # The array of sign in guards to run when signing a user in.
+    # Defaults to an empty array. Sign in guards respond to `call` and are
+    # initialized with a session and the current stack. Each guard can decide
+    # to fail the sign in, yield to the next guard, or allow the sign in.
+    # @return [Array #call]
```
mcmire: What does `Array #call` mean here? (Just curious)

.yardopts
```diff
@@ -0,0 +1,3 @@
+--exclude templates
+--markup markdown
+--markup-provider redcarpet
```
mcmire: You might also want to include:

```
--no-private
--protected
```

lib/clearance/configuration.rb
derekprior: I'm trying to say it's an array of things that respond to `#call`. But it doesn't actually seem to work the way I'm lead to believe it works here: http://yardoc.org/types.html

mcmire:
I am not a YARD expert, I just use a few of the features, but other than the comment above, everything seems to look good here.

lib/clearance/configuration.rb
gabebw: Maybe you want `Array<#call>`?

PR #528, opened 2015-01-30
Enable forgery protection on sessions#create

derekprior:
This line has existed since 2008, and yet I can determine no justification for
it. It seems to me that we *would* want CSRF protection on `session#create`.

On its own, skipping CSRF protection in just this single action doesn't seem
particularly useful to an attacker. Additional vectors (such as an
overly-permissive CORS header) would have to be present to make use of this,
but at that point far more interesting attacks would be possible on any
cookie-based auth system.


gabebw:
I am not particularly security minded.

That said, this is a Rails default and we don't NEED to turn it off for anything we do. If someone's relying on this, that probably points to a security flaw in their app, right?

sgrif:
:+1: If I had to guess, someone was hitting this with JavaScript and forgot to or didn't know how to send the CSRF token along. By skipping it here, we open ourselves up to potential XSS vulnerabilities and/or clickjacking attacks.

PR #530, opened 2015-01-31
Don't expose authentication methods as actions

derekprior:
There were a couple of methods from the Clearance::Authentication that were
leaking through to become routable action methods on the controller. We need
to hide them as actions.


jferris:
Think it's worth having a test for this? My idea:

* Make an anonymous controller subclass using rspec-rails.
* Mix in `Clearance::Controller`.
* Make sure there are no routable actions.

derekprior:
That makes a lot of sense.

PR #532, opened 2015-02-06
Set an expiration for password reset tokens

### LINKED ISSUE #465: Expire password reset token after a period of time

mxie:
Overview:
Password reset tokens will now have an expiration date/time. By default, it
will be in 15 minutes. This is configurable via the `password_reset_time_limit`
option.

Users can request a reset as many times as they wish. Once they "claim" a reset
with one of their tokens, all others will be invalidated (set as expired).

Details:
* User-token data in a separate table called `password_resets`.
* Password reset logic has been extracted from the `User` model. It's now
handled at the controller level, referring to the `PasswordReset` and
`PasswordResetDeactivator` classes as needed.
* Routes and views have stayed the same for transparency.
  - To consider: renaming `PasswordsController` to `PasswordResetsController`
  later.
* Introduced `Clearance::Generators::Migration` to abstract common migration-related methods

This has been tested on a repo that's currently using Clearance, and it seems to work fine. I plan on adding some info on the Wiki for anyone who's adding the expiration feature to their existing Clearance'd project.

Also, this is a huge PR. Suggestions on how to break this up would be great appreciated.

Fixes #465


spec/generators/clearance/password_reset_migration/password_reset_migration_generator_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "spec_helper"
+require "generators/clearance/password_reset_migration/password_reset_migration_generator"
```
mxie: We could either ignore this or rename this generator. I have no particular preference.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -5,17 +5,20 @@ class Clearance::PasswordsController < Clearance::BaseController
...
   before_filter :forbid_non_existent_user, only: [:edit, :update]
+  before_filter :forbid_expired_password_reset, only: [:edit, :update]
 
   def create
     if user = find_user_for_create
-      user.forgot_password!
-      deliver_email(user)
+      password_reset = PasswordReset.new
+      password_reset.user = user
+      password_reset.save
```
mxie: It's been a while since I've Rails'ed, but there's a better way to do this, right? It's a one-way association, i.e. there's only a `belongs_to` association in `PasswordReset`, no `has_many` from `User`.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -24,10 +27,12 @@ def new
...
 
   def update
-    @user = find_user_for_update
+    @password_reset = find_password_reset_by_user_id_and_token
+    user = @password_reset.user
 
-    if @user.update_password password_reset_params
-      sign_in @user
+    if user.update_password password_reset_params
+      Clearance::PasswordResetDeactivator.new(user).run
```
mxie: I don't know if this is excessive. It feels right since the responsibility is separated, but the `run` call is really only calling `PasswordReset.active_for(user).deactivate_all`, which could be easily used here instead.

mike-burns:
Since this changes the interface for documented template methods (the documentation is about as broad as possible: "See app/controllers/clearance for the default behavior"), this means that this will be part of the major version bump. Right?

mxie:
@mike-burns Yes, you are correct.

app/controllers/clearance/passwords_controller.rb
derekprior: Maybe:

```ruby
password_reset = PasswordReset.create!(user: user)
```

app/controllers/clearance/passwords_controller.rb
derekprior: This probably needs to happen in a transaction as well right? What happens if the password change succeeds but the deactivation does not?

db/migrate/20150122204353_create_password_resets.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePasswordResets < ActiveRecord::Migration
+  def change
+    create_table :password_resets do |t|
+      t.integer :user_id, null: false
+      t.string :token, limit: 128, null: false
+      t.datetime :expires_at, null: false
+      t.timestamps null: false
+    end
+
+    add_index :password_resets, :user_id
```
derekprior: should we index the `expires_at` column as well? Maybe change this to a compound index on `[:user_id, :expires_at]`? Might be too much...

db/migrate/20150127232904_remove_confirmation_token_from_users.rb
```diff
@@ -0,0 +1,33 @@
...
+        (user_id, token, expires_at, created_at, updated_at)
+      SELECT
+        users.id,
+        users.confirmation_token,
+        '#{expiration_timestamp}',
+        CURRENT_TIMESTAMP,
+        CURRENT_TIMESTAMP
+      FROM users
+      WHERE users.confirmation_token IS NOT NULL
+    SQL
```
derekprior: good thinking!

spec/generators/clearance/password_reset_migration/password_reset_migration_generator_spec.rb
derekprior: ignore.

derekprior:
Took a quick spin through everything. This is great, thanks so much.

I do think we can do this in a manner that lets us ship this in opt-in fashion for 1.x. Either via a separate controller and/or via service objects like monban. We can look at it Friday.

app/controllers/clearance/passwords_controller.rb
mxie: Yeah, I had something like that originally, but Rails 3.2 doesn't seem to like it because of mass-assignments. =\ I guess I could at least extract the reset-building as a private method in here.

app/controllers/clearance/passwords_controller.rb
mxie: Hmm...we wouldn't want the resets to successfully deactivate if the password update failed, right? We could check for successful runs for both via:

```ruby
if user.update_password "whatever" && deactivator.run
``` 

...or something like that.

Was there some other flow you were thinking of?

app/controllers/clearance/passwords_controller.rb
mxie: Did something like this instead: f2ee26c.

db/migrate/20150122204353_create_password_resets.rb
mxie: Sounds good to me. Done in 8f0f9a8.

app/models/password_reset.rb
```diff
@@ -0,0 +1,36 @@
+class PasswordReset < ActiveRecord::Base
```
mxie: Should this be a template instead of in `app/models`, like `lib/generators/clearance/install/templates/user.rb`?

derekprior:
I have not forgotten about this, I've just been out or otherwise engaged on investment days. I'll be out this Friday as well but will try to spend some time playing with this on the plane.

mxie:
@derekprior Not a problem. I've been working on a bunch of other stuff in the meantime anyway. I actually haven't gotten to making this feature an opt-in one yet, so no rush.

README.md
```diff
@@ -106,15 +110,30 @@ helpers. For example:
 
 ### Password Resets
 
-When a user resets their password, Clearance delivers them an email. You
-should change the `mailer_sender` default, used in the email's "from" header:
+When a user resets their password, Clearance delivers them an email. By default,
+the password reset token expires in 15 minutes. You can change the time limit by
+passing in an `ActiveSupport::Duration`. You should also change the
```
jessieay: > You can change the time limit by passing in an `ActiveSupport::Duration` to `config.password_reset_time_limit`

lib/clearance/configuration.rb
```diff
@@ -93,6 +98,7 @@ def initialize
       @cookie_name = "remember_token"
       @httponly = false
       @mailer_sender = 'reply@example.com'
+      @password_reset_time_limit = 15.minutes
```
jessieay: `ENV` var for this?

mxie: @jessieay - I don't think I understand why this should be an `ENV` var. I also don't think it'll work as one either since the environment wouldn't understand `15.minutes`. Can you further elaborate?

README.md
mxie: Thanks! Changed in b101536.

lib/clearance/configuration.rb
jessieay: you could make `15` the env var and then do `ENV["PASSWORD_RESET_TIME_IN_MINUTES"].to_i.minutes` 

I am not 100% sure it makes sense here, but my general rule in a regular rails app is that time-related variables should be `ENV` vars in case they need to be changed on the fly. For example, on my current project, we have a `time_to_refresh` var that is 15 minutes on staging and prod for performance reasons, but sometimes we want to move it down to 1 minute of we are testing something on staging.

Your call if you want to go that route here, though!

derekprior: Clearance users are free to configure this to use an environment variable in their initializer, but we have to provide a valid default from the gem.

README.md
```diff
@@ -106,15 +110,34 @@ helpers. For example:
 
 ### Password Resets
 
-When a user resets their password, Clearance delivers them an email. You
-should change the `mailer_sender` default, used in the email's "from" header:
+When a user resets their password, Clearance delivers them an email.
```
derekprior: How about "Clearance sends them an email."

app/controllers/clearance/passwords_controller.rb
```diff
@@ -37,8 +40,14 @@ def update
 
   private
 
-  def deliver_email(user)
-    mail = ::ClearanceMailer.change_password(user)
+  def create_password_reset_for(user)
```
derekprior: SInce we're going to target this for 2.0, I think we should not worry about attr_accessible and use strong params here. What do you think?

derekprior: SInce we're going to target this for 2.0, I think we should not worry about attr_accessible and use strong params here. What do you think?

app/controllers/clearance/passwords_controller.rb
```diff
@@ -37,8 +40,14 @@ def update
...
+    PasswordReset.create! do |password_reset|
```
derekprior: SInce we're going to target this for 2.0, I think we should not worry about attr_accessible and use strong params here. What do you think?

app/controllers/clearance/passwords_controller.rb
```diff
@@ -56,11 +65,15 @@ def password_reset_params
     end
   end
 
-  def find_user_by_id_and_confirmation_token
-    user_param = Clearance.configuration.user_id_parameter
+  def find_password_reset_by_user_id_and_token
```
derekprior: What do you think about naming this `find_password_reset` instead? I think it might be a more override-friendly name. For instance, if you add some functionality and need to find the password reset by it as well, this method name would be misleading.

app/controllers/clearance/passwords_controller.rb
derekprior: I was thinking of a scenario where the password gets updated but then for whatever reason the deactivator errors out. In that case, their password would be changed but their token would still be valid. To get around that, you can:

```ruby
ActiveRecord::Base.transaction do
  user.update_password password_reset_params
  deactivator.run
  true
end
```

In that way, if any of those statements fail, the database changes are rolled back.

derekprior: It may make sense to do something like this in the model:

```ruby
def complete(params)
  transaction do
    user.update_password(params)
    deactivate_user_resets
    true
  end
end
```
Then the controller can fire off this complete method and inspect the result to make sure everything worked.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -96,12 +101,27 @@ def forbid_missing_token
...
     end
   end
 
+  def find_user_from_password_reset
+    if matching_password_reset = find_password_reset_by_user_id_and_token
+      matching_password_reset.user
+    end
+  end
+
+  def forbid_expired_password_reset
```
derekprior: would this fit as a validation on password reset instead? validate it's not expired, call `valid?` as part of the theoretical `complete` method, perhaps?

app/models/password_reset.rb
derekprior: No. Clearance models are internal to clearance and not generate. User is a template because we expect you already have a user class of some sort.

app/controllers/clearance/passwords_controller.rb
mxie: Hm, I don't think we want to use strong params here. IIRC, this was done so that we could associate the password reset with a user. I don't recall why I went for a block here instead of just `PasswordReset.create!(user: user)` though...so I'll just go ahead and simplify it to that.

db/migrate/20150122204353_create_password_resets.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePasswordResets < ActiveRecord::Migration
+  def change
+    create_table :password_resets do |t|
+      t.integer :user_id, null: false
+      t.string :token, limit: 128, null: false
```
derekprior: do you know what the significance of the limit is here?

app/controllers/clearance/passwords_controller.rb
mxie: I agree! I think I was just trying to follow the same naming convention as it was previously.

lib/clearance/configuration.rb
derekprior: @mxie Did you base the default on anything other libraries are doing for this? I'm wondering if it might be a bit too low, but I don't really know. I know that when I have to reset my password, sometimes the email doesn't come immediately and I get distracted doing something else. Maybe `1.hour`? I don't feel strongly and can easily be convinced 15 minutes is correct.

mxie: According to the OWASP link from the GH issue (http://owasp.com/index.php/Forgot_Password_Cheat_Sheet#Step_3.29_Send_a_Token_Over_a_Side-Channel), they’re saying no more than 20. I think I remembered that somehow and kind of picked 15 arbitrarily.


On June 5, 2015 at 10:39:57 AM, Derek Prior (notifications@github.com) wrote:

In lib/clearance/configuration.rb:

> @@ -93,6 +98,7 @@ def initialize
>        @cookie_name = "remember_token"
>        @httponly = false
>        @mailer_sender = 'reply@example.com'
> +      @password_reset_time_limit = 15.minutes
@mxie Did you base the default on anything other libraries are doing for this? I'm wondering if it might be a bit too low, but I don't really know. I know that when I have to reset my password, sometimes the email doesn't come immediately and I get distracted doing something else. Maybe 1.hour? I don't feel strongly and can easily be convinced 15 minutes is correct.

derekprior:
Ok, sorry it took me so long to get back to this. I like this a lot. I have a few questions or suggestions. Let me know what you think. I'll take a look at the state of master and decide whether the merge target for this should be master (that is I expect to release 2.0 soonish) or if I want it to go to a 2.0 branch while I deprecate a few more things in 1.x...

lib/clearance/configuration.rb
derekprior: Nice. Let's go with 20 then. I'm happy to follow OWASP's upper boundry here.

db/migrate/20150122204353_create_password_resets.rb
mxie: I believe this limit was set for the confirmation_token previously. I don’t know what the rationale behind it was though. It’s also interesting because we’re limiting the token to a length of 20 bytes, so they don’t seem to match up. Thoughts on removing the limit?


On June 5, 2015 at 10:37:52 AM, Derek Prior (notifications@github.com) wrote:

In db/migrate/20150122204353_create_password_resets.rb:

> @@ -0,0 +1,12 @@
> +class CreatePasswordResets < ActiveRecord::Migration
> +  def change
> +    create_table :password_resets do |t|
> +      t.integer :user_id, null: false
> +      t.string :token, limit: 128, null: false
do you know what the significance of the limit is here?

mike-burns: http://blog.jonanin.com/2013/11/20/postgresql-char-varchar/

This forces a VARCHAR where it could instead be TEXT. +1 to removing the limit.

app/controllers/clearance/passwords_controller.rb
mxie: Great idea! I've actually never used ActiveRecord Transactions before, and this totally makes sense. I'll give it a go.

db/migrate/20150122204353_create_password_resets.rb
mxie: Done in df5083e.

app/controllers/clearance/passwords_controller.rb
mxie: Done in 948a4f7

app/controllers/clearance/passwords_controller.rb
mxie: Done in 48a3a4c

README.md
mxie: Done in 789933b

app/controllers/clearance/passwords_controller.rb
mxie: I did some crazy stuff in https://github.com/thoughtbot/clearance/commit/49ae7ac3d029b2250d37be8ded3b94c5d89a11cb. Let me know what your thoughts are.

I'm finding that validating that the password update is valid kind of hard when wrapped in a transaction...might need some more brainstorming here where we can use `ActiveRecord::Base.transaction` but still be able to fetch useful error messages from either the user or the password reset objects.

mxie:
@derekprior Thanks for taking another look! I made a bunch of changes based on your comments and suggestions. There are some parts that might still require some brainstorming though.

If there are a few other things you'd like to fix, I think a 2.0 branch makes the most sense. I don't think there's a huge urgency for this feature at the moment anyway.

mxie:
@derekprior Whenever you get a chance, could you take a look at this again? I think it's pretty much all set to go. The only thing I might add is an "Upgrade Guide" of sorts. That could be done separately though.

derekprior:
With Clearance 2.0 progressing, I was finally able to return to this. I made some major changes that are afforded by dropping support for older versions of Rails. We now no longer need the database for password reset tokens, so the resulting PR is simpler, I think.

See #682.

PR #533, opened 2015-02-10
Update to latest shoulda-matchers

derekprior:
Change references to deprecated `set_the_flash` matcher.


mxie:
There seems to be a test failure on Travis CI:

```
1) Clearance::RackSession injects a clearance session into the environment
    Failure/Error: expect(Clearance::Session).to have_received(:new).with(env)
      `have_received(...).with(...)` cannot be used when received message args have later been mutated. You can use a normal message expectation (`expect(...).to receive(...).with(...)`) instead.
```

Changes look fine to me otherwise though,

Issue #531, opened 2015-02-02
Requiring clearance rspec helper breaks rspec.


txdavidtx:
When I add the require statement ```require "clearance/rspec"``` to use ```Clearance::BackDoor``` it breaks rspec and gives the following error:
```
/.rvm/gems/ruby-2.1.1/gems/activesupport-4.2.0/lib/active_support/core_ext/module/deprecation.rb:21:in `deprecate': uninitialized constant Module::ActiveSupport (NameError)
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/activesupport-4.2.0/lib/active_support/core_ext/class/delegating_attributes.rb:26:in `<class:Class>'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/activesupport-4.2.0/lib/active_support/core_ext/class/delegating_attributes.rb:6:in `<top (required)>'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/activesupport-4.2.0/lib/active_support/core_ext/class.rb:2:in `require'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/activesupport-4.2.0/lib/active_support/core_ext/class.rb:2:in `<top (required)>'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/actionview-4.2.0/lib/action_view/template/resolver.rb:2:in `require'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/actionview-4.2.0/lib/action_view/template/resolver.rb:2:in `<top (required)>'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/actionview-4.2.0/lib/action_view/testing/resolvers.rb:1:in `require'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/actionview-4.2.0/lib/action_view/testing/resolvers.rb:1:in `<top (required)>'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/rspec-rails-3.1.0/lib/rspec/rails/view_rendering.rb:1:in `require'
	from /Users/davidhunt/.rvm/gems/ruby-2.1.1/gems/rspec-rails-3.1.0/lib/rspec/rails/view_rendering.rb:1:in `<top (required)>'
[...]

PR #534, opened 2015-02-17
Update README to reflect changes in RSpec 3

pedrosmmoreira:
Adding `clearance/rspec` to `spec/spec_helper.rb` in RSpec versions
greater or equal to 3 will throw an `uninitialized constant
Module::ActiveSupport` error, since the Rails env is not available.

This updates the README in order to indicate the require statement
should be placed in `spec/rails_helper.rb`.

Adresses issue https://github.com/thoughtbot/clearance/issues/531

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,11 +339,15 @@ $ rails generate clearance:specs
 To test controller actions that are protected by `before_filter :require_login`,
 require Clearance's test helpers in your test suite.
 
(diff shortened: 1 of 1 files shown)


derekprior:
I tried to shorten it a bit and bias it towards RSpec 3+. Merged in 8141301. Thanks so much for this!

PR #535, opened 2015-03-02
Update rspec-rails version

mxie:
As a result:
* In the ActionMailer initializer for the test app, we have to refer to the
ActionMailer object itself to configure it instead of the app config. This was
recommended in https://github.com/rspec/rspec-rails/issues/1313.
* We had to change the setup for rack_session_spec because RSpec now checks for
mutation of arguments passed into spies.


derekprior:
:beers:

PR #537, opened 2015-03-03
Upgrade shoulda-matchers for Rails 3.2 testing too

mxie:


derekprior:
ah yes. I always forget to switch rubies...

PR #540, opened 2015-03-12
Mention coaching

jferris:


derekprior:
Seems fine. I'd also be okay with this above the "Contributing" section if that would help. Also maybe worth adding to the final line of the readme? Something like:

> We are available for hire and 1-on-1 coaching?

jferris:
@derekprior I think "Contributing" is more core to open source software, so I left it first. I took a shot at working it into the "About" section. Thoughts?

derekprior:
:+1:

Issue #541, opened 2015-03-14
Clearance in development ignores updates on User model


sanderhahn:
The User model that clearance uses is configured during startup in the initialiser. During development changes on your User model or on relationships are ignored which might not be as expected. The Rails server has to be explicitly restarted to get objects of the updated class(es).

The Spring documentation (https://github.com/rails/spring) under Class Reloading says: "Don't save off references to application constants in your initialization code." 

One workaround i found is to explicitly reload the clearance initializer in a to_prepare callback:

```ruby
# config/initializers/clearance.rb
class Railtie < ::Rails::Railtie
  config.to_prepare do
    load __FILE__
  end
end

Clearance.configure do |config|
  config.mailer_sender = "reply@example.com"
  config.user_model = User
end
```

Thanks for the clearance gem :-)
Best regards,
Sander

derekprior:
Ugh, Spring.

There isn't a great way to configure a user class without storing a reference to that class. I could perhaps take a lambda instead but `config.user_model = -> { User }` is a pretty nonsensical configuration that would exist only to solve a problem with pre-loaders.

I could generate an initializer like you have above but that seems like it would be super confusing to most people. I don't see a good solution here at the moment. What do you think?

derekprior:
Hmmm. Wonder if I could change the configuration reader to do something like:

```ruby
def user_model
  -> { @user_model }.call
end
```

Or perhaps only do that in the case where we're running with a pre-loader or something?

derekprior:
I assume that's still going to have the same problem.

sanderhahn:
Maybe its possible to add a Railtie `to_prepare` callback inside clearance for `Rails.env.development?` to relookup the user class constant when it is defined? That might fix it without adjust the api of the configuration.

derekprior:
I think that means I'd have to hard-code `config/initializers/clearance.rb` as the file path to reload. It's an option.

sanderhahn:
How about adding to `lib/clearance/configuration.rb`:

```ruby
  # Relookup user_model on Rails restart
  class MyRailtie < Rails::Railtie
    config.to_prepare do
      Clearance.configuration.user_model = Clearance.configuration.user_model.to_s.constantize
    end
  end
```

sanderhahn:
Sorry about all the references, my proposed solution also doesn't seem to work when i run the tests. Will see if i can fix it...

lenart:
[...]

PR #543, opened 2015-03-20
Reload user_model in Railtie, proposal fix for #541

sanderhahn:


sanderhahn:
Sorry this was not a good solution...

PR #544, opened 2015-03-21
Reload user_model in Railtie, proposal fix for #541

sanderhahn:
One more try...


derekprior:
Cleaned up a bit in #547

PR #546, opened 2015-03-26
added 'Change my password' link to password mail template

farukaydin:
It solves rubygems/rubygems.org#908 problem.

### DIFF
--- a/app/views/clearance_mailer/change_password.html.erb
+++ b/app/views/clearance_mailer/change_password.html.erb
@@ -1,5 +1,5 @@
-<%= t(".opening") %>
+<p><%= t(".opening") %></p>
 
-<%= edit_user_password_url(@user, token: @user.confirmation_token.html_safe) %>
+<p><%= link_to 'Change my password', edit_user_password_url(@user, token: @user.confirmation_token.html_safe) %></p>
 
-<%= raw t(".closing") %>
+<p><%= raw t(".closing") %></p>
--- a/app/views/clearance_mailer/change_password.html.erb
+++ b/app/views/clearance_mailer/change_password.html.erb
@@ -1,5 +1,5 @@
 <p><%= t(".opening") %></p>
 
-<p><%= link_to 'Change my password', edit_user_password_url(@user, token: @user.confirmation_token.html_safe) %></p>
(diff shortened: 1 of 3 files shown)


app/views/clearance_mailer/change_password.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= t(".opening") %>
+<p><%= t(".opening") %></p>
 
-<%= edit_user_password_url(@user, token: @user.confirmation_token.html_safe) %>
+<p><%= link_to 'Change my password', edit_user_password_url(@user, token: @user.confirmation_token.html_safe) %></p>
```
jessieay: Can we use a locale for this text? (the rest of the email text is from I18n so that seems like it would be good to stay consistent)

spec/mailers/clearance_mailer_spec.rb
```diff
@@ -29,6 +29,7 @@
     email = ClearanceMailer.change_password(user)
 
     expect(email.body.to_s).to include(link)
+    expect(email.body.to_s).to have_css("a", text: "Change my password")
```
jessieay: let's use I18n in the expectation when the template is updated

farukaydin:
@jessieay updated pr to use i18n.

yes rubygems uses clearance :sparkles:

jessieay:
@farukaydin awesome, thanks for the update! This looks good to me. I will let @tute have the final say on this before merging.

jessieay:
And by @tute, I mean @derekprior :space_invader: forgot which library we were in here! :)

derekprior:
Introducing a new key for existing text might be tricky to get right. If your language is set to french, for instance, you don't get the benefit of the default translations we provide. So now your password reset mailer has a link with the text "link text". We can't rely on apps using clearance to have configured english as a fallback.

We may need to provide a `default: Change your password` option to the translate call. Thoughts?

derekprior:
Updated and merged in 41f06bef59563cad535a339431a1b7f5623c6e44

derekprior:
Thank you @farukaydin.

farukaydin:
@derekprior you're right, default option looks good :+1:

PR #547, opened 2015-03-27
Allow User model to be reloaded in development

derekprior:
When a `user_model` is configured in a Clearance initializer, a reference
to that class is immediately saved off. If that class is changed, Clearance
will not know to automatically reload the class as Rails does automatically
for classes in development.

This change introduces a `to_prepare` block to the Engine that is
responsible for forcing the configured user class to be reloaded.
`to_prepare` runs once per request in development and only at startup in
other environments.

### DIFF
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -133,6 +133,17 @@ module Clearance
     def routes_enabled?
       @routes
     end
+
+    # Reloads the clearance user model class.
+    # This is called from the Clearance engine to reload the configured
+    # user class during each request while in development mode, but only once
+    # in production.
+    # @private
+    def reload_user_model
+      if @user_model.present?
+        @user_model = @user_model.to_s.constantize
+      end
+    end
   end
 
   # @return [Clearance::Configuration] Clearance's current configuration
--- a/lib/clearance/engine.rb
+++ b/lib/clearance/engine.rb
@@ -1,12 +1,15 @@
-require 'clearance'
-require 'rails'
+require "clearance"
+require "rails"
 
 module Clearance
   class Engine < Rails::Engine
-    initializer 'clearance.filter' do |app|
+    initializer "clearance.filter" do |app|
       app.config.filter_parameters += [:password, :token]
     end
 
-    config.app_middleware.insert_after ActionDispatch::ParamsParser, Clearance::RackSession
+    config.app_middleware.insert_after(
+      ActionDispatch::ParamsParser,
+      Clearance::RackSession
+    )
   end
 end
--- a/lib/clearance/engine.rb
+++ b/lib/clearance/engine.rb
@@ -11,5 +11,9 @@ module Clearance
       ActionDispatch::ParamsParser,
       Clearance::RackSession
     )
+
+    config.to_prepare do
+      Clearance.configuration.reload_user_model
+    end
   end
 end
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -162,4 +162,19 @@ describe Clearance::Configuration do
       expect(Clearance.configuration.routes_enabled?).to be false
     end
   end
+
+  describe "#reload_user_model" do
+    it "returns the user model class if one has already been configured" do
+      ConfiguredUser = Class.new
+      Clearance.configure { |config| config.user_model = ConfiguredUser }
+
+      expect(Clearance.configuration.reload_user_model).to eq ConfiguredUser
+    end
+
+    it "returns nil if the user_model has not been configured" do
+      Clearance.configuration = Clearance::Configuration.new
+
+      expect(Clearance.configuration.reload_user_model).to be_nil
+    end
+  end
 end


lib/clearance/configuration.rb
```diff
@@ -133,6 +133,16 @@ def user_id_parameter
...
       @routes
     end
+
+    # Reloads the clearance user model class.
+    # This is called from the Clearance engine to reload the configured
+    # user class during each request while in development mode.
+    # @private
+    def reload_user_model
+      if @user_model.present?
+        @user_model = @user_model.to_s.constantize
```
jessieay:
```
 pry(main)> ::User.to_s.constantize
=> User (call 'User.connection' to establish a connection)
::User
=> User (call 'User.connection' to establish a connection)
```

since these return the same value, do we need to `to_s` and `constantize`?

lib/clearance/engine.rb
```diff
@@ -1,12 +1,19 @@
...
   class Engine < Rails::Engine
-    initializer 'clearance.filter' do |app|
+    initializer "clearance.filter" do |app|
       app.config.filter_parameters += [:password, :token]
     end
 
-    config.app_middleware.insert_after ActionDispatch::ParamsParser, Clearance::RackSession
+    config.app_middleware.insert_after(
+      ActionDispatch::ParamsParser,
+      Clearance::RackSession
```
jessieay: we doing trailing commas in this library?

lib/clearance/configuration.rb
```diff
@@ -133,6 +133,16 @@ def user_id_parameter
     def routes_enabled?
       @routes
     end
+
+    # Reloads the clearance user model class.
+    # This is called from the Clearance engine to reload the configured
+    # user class during each request while in development mode.
```
jessieay: The comment in the commit msg about this happening only on startup in non-development environments was interesting / informative. Should we include that info here as well?

lib/clearance/configuration.rb
derekprior: Yes, we basically have to force Ruby to re-look-up the constant, which is what we accomplish with `constantize` from a string. In an actual application running in development mode, if you make changes to the user class, the reference saved by clearance won't actually be the same class that you get with this constantize dance.

 As an added benefit (perhaps) this also is more permissive in configuration as users could pass a string or symbol which would still be constantized properly.

sanderhahn:
Nice! :)

Nitpicking: Rails::Railtie::Configuration#to_prepare defines generic callbacks to run before #after_initialize (not necessarily on every request). The ActionDispatch::Reloader#to_prepare callbacks are run before each request.

lib/clearance/configuration.rb
jessieay: Thanks for the explanation.

derekprior:
> The ActionDispatch::Reloader#to_prepare callbacks are run before each request.

That seems to be the context we're in inside that `to_prepare` block for the Engine.

lib/clearance/engine.rb
derekprior: I wasn't even aware you could have trailing commas in method calls. I tried it out and it does work, but I don't think I've seen that convention outside of arrays and hashes.

lib/clearance/configuration.rb
derekprior: added.

lib/clearance/engine.rb
jessieay: My current client uses Rubocop in their codebase and the conventions they have enforce trailing commas in hashes, arrays, and method args. But to each his own!

sanderhahn:
You are calling the `config.to_prepare` in a class extending the Railtie::Engine. That would be: http://api.rubyonrails.org/classes/Rails/Railtie/Configuration.html#method-i-to_prepare. When i was experimenting i found that the reload triggered on every model class change but not necessarily on every request.

sanderhahn:
Was not sure how to test if reloading was working and looked over the engine. This spec tests if the callback has been configured, which might be enough if you can rely on Rails to properly call them:

```ruby
  describe "#reload_user_model" do
    it "expect reload_user_model to be called on framework reload" do
      expect(Clearance::Engine.config.to_prepare_blocks).to include(&:reload_user_model)
    end
  end
```

lib/clearance/engine.rb
sanderhahn: Would be nice to see a pretty printer/formatter for ruby that makes your code formatted in the one true way according to a certain style guide ;-)

derekprior: @sanderhahn rubocop can re-write your code according to rules if you'd like: https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/corrector.rb

derekprior:
I thought about that too. At that point I'm just guarding against someone deleting the `to_prepare` block. I wouldn't be testing actual behavior. I'm hopeful people will read documentation or git history for the impacted lines and think that's higher value than this particularl test.

derekprior:
It seems my tests for `reload_user_model` are non-deterministic, so I'll need to spend a bit more time here.

lib/clearance/engine.rb
sanderhahn: Great, didn't know that thanks :) Will try it out!

PR #548, opened 2015-03-31
Correct unnecessary spacing in migration file

pedrosmmoreira:
This is a small correction to stop tools such as HoundCi throwing an
`Unnecessary spacing detected` comment.

While a very minor change, it is one less "fix and rebase" step during code
review.


derekprior:
Thanks!

PR #550, opened 2015-04-03
Include a text part for change password mailer

derekprior:
It's best practice to send a text part when sending HTML email. It's
relatively simple for us to do so here even at the cost of some
duplication.


kenyonj:
LGTM ... other than the failing CI

PR #552, opened 2015-04-04
Use the same syntax for all translations in the views

l3kn:
The syntax for translations is inconsistent between different views and mail templates.

It might be that I am unaware of some style guideline regarding this, but using `t('.title')` seems to be a good compromise.


derekprior:
I'd file this as style, rather than syntax, but it bothers me that they aren't consistent as well. We've been transitioning from single to double quotes with changes, though, so if you want to switch those to double quotes, I'll merge this in.

PR #553, opened 2015-04-05
Use the same style for all translations in the views

l3kn:
(reopened with fixed branch name & commit messages)

The style for translations is inconsistent between different views and mail templates.

It might be that I am unaware of some style guideline regarding this, but using t(".title") seems to be a good compromise.


derekprior:
merged in cc65610, thanks!

PR #555, opened 2015-04-15
Clarify overriding controllers

paulcsmith:
It does not mention needing to specify a new route, but that is
required.


derekprior:
:+1:

PR #558, opened 2015-05-08
Deprecate unused password strategies

derekprior:
We don't use `SHA1` or `Blowfish` password strategies. I have no plans
to maintain them moving forward and as an opinionated library Clearance
should be pushing users to `BCrypt`. With `SHA1` being deprecated we
also need to deprecated the `BCryptMigrationFromSHA1` strategy.

These strategies can still be used, deprecation free, by adding the
`clearance-deprecated_password_strategies` gem. We do not intend to
maintain these, but also have no plans to break the interface. Should
maintainers want to keep a particular strategy alive and maintained,
they should extract it.

### DIFF
--- a/lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
+++ b/lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
@@ -1,6 +1,13 @@
 module Clearance
   module PasswordStrategies
     module BCryptMigrationFromSHA1
+      DEPRECATION_MESSAGE = "[DEPRECATION] The BCryptMigrationFromSha1 " \
+        "password strategy has been deprecated and will be removed from " \
+        "Clearance 2.0. BCrypt is the only officially supported strategy, " \
+        "though you are free to provide your own. To continue using this " \
+        "strategy, add clearance-deprecated_password_strategies to your " \
+        "Gemfile."
+
       class BCryptUser
         include Clearance::PasswordStrategies::BCrypt
 
@@ -22,10 +29,12 @@ module Clearance
       end
 
       def authenticated?(password)
+        warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
         authenticated_with_sha1?(password) || authenticated_with_bcrypt?(password)
       end
 
       def password=(new_password)
+        warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
         @password = new_password
         BCryptUser.new(self).password = new_password
       end
--- a/lib/clearance/password_strategies/sha1.rb
+++ b/lib/clearance/password_strategies/sha1.rb
@@ -3,13 +3,21 @@ module Clearance
     module SHA1
       require 'digest/sha1'
 
+      DEPRECATION_MESSAGE = "[DEPRECATION] The SHA1 password strategy " \
+        "has been deprecated and will be removed from Clearance 2.0. BCrypt " \
+        "is the only officially supported strategy, though you are free to " \
+        "provide your own. To continue using this strategy add " \
+        "clearance-deprecated_password_strategies to your Gemfile."
+
       extend ActiveSupport::Concern
 
       def authenticated?(password)
+        warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
         encrypted_password == encrypt(password)
       end
 
       def password=(new_password)
+        warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
         @password = new_password
         initialize_salt_if_necessary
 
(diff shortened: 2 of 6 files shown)


derekprior:
The referenced gem is not live yet, but is available to other thoughtbotters for review at https://github.com/thoughtbot/clearance-deprecated_password_strategies

lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
```diff
@@ -1,6 +1,13 @@
 module Clearance
   module PasswordStrategies
     module BCryptMigrationFromSHA1
+      DEPRECATION_MESSAGE = "[DEPRECATION] The BCryptMigrationFromSha1 " \
```
mcmire: I don't know if there are any other warnings in this project, but would it be a good idea to use a heredoc and wrap this at 72 or 80 characters?

lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
```diff
@@ -22,10 +29,12 @@ def initialize(user)
       end
 
       def authenticated?(password)
+        warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
```
mcmire: Would it be a good idea to extract all of these warnings to some kind of `warn` utility method which would take care of prepending the message with the caller?

mcmire:
Two suggestions, but otherwise LGTM

lib/clearance/password_strategies/sha1.rb
```diff
@@ -3,13 +3,21 @@ module PasswordStrategies
     module SHA1
       require 'digest/sha1'
 
+      DEPRECATION_MESSAGE = "[DEPRECATION] The SHA1 password strategy " \
```
jessieay: this seems like almost an exact dupe of the message in `blowfish.rb` - is there somewhere you can put this constant where the two can share it?

lib/clearance/password_strategies/sha1.rb
```diff
@@ -3,13 +3,21 @@ module PasswordStrategies
...
+        "has been deprecated and will be removed from Clearance 2.0. BCrypt " \
+        "is the only officially supported strategy, though you are free to " \
+        "provide your own. To continue using this strategy add " \
+        "clearance-deprecated_password_strategies to your Gemfile."
```
jessieay: this is a little thing, but is there a reason why this gem has dashes and underscores in the name? Seems unusual.

derekprior: Common convention is that `-` denotes an extension to whatever precedes it, while `_` is the usual word separator. So since this is a clearance extension, it should be `clearance-` and then follow our usual rules. `factory_girl_rails` should actually be `factory_girl-rails`. I agree that it's ugly, but that's the guide. See: http://guides.rubygems.org/name-your-gem/

lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
derekprior: If I'm not mistaken, the choice is between a heredoc and proper indentation? I'd rather have the indentation. Any other options?

derekprior: To elaborate - if I use a heredoc, I think I have to leaves the messages left aligned in the source or I'll actually get those spaces in the terminal output. Is that correct?

lib/clearance/password_strategies/sha1.rb
derekprior: Yeah, I noticed this as well. I thought about it, but the "shared place" for that method to live would be the `PasswordStrategies` module which means adding code to present deprecations to parts of the app that are not being deprecated. I think I'm okay with the duplication here since it's never going to change once committed until the files are removed in 2.0.

lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
derekprior: If I'm not mistaken, the choice is between a heredoc and proper indentation? I'd rather have the indentation. Any other options?

If I use a heredoc, I think I have to leaves the messages left aligned in the source or I'll actually get those spaces in the terminal output. Is that correct?

lib/clearance/password_strategies/sha1.rb
jferris: FYI, it was originally `factory_girl-rails` but enough people complained that I renamed it.

derekprior: Would you recommend bucking the convention here?

jferris: No, I don't think so. People may be used to the convention enough now that they won't complain.

PR #563, opened 2015-06-11
Add support for multiple base domains

agraves:
This PR adds support for a Rails app that serves multiple distinct FQDNs. In other words, one user may log into `client.foo.com` and another may use `other_client.bar.com`. This is not an attempt to allow users to log into `client.foo.com` and then access pages on `other_client.bar.com` -- but rather, to ensure that a single app can issue cookies for multiple base domains.

I recognize that multiple base domains is an undesirable pattern, but it's a reality for many Rails apps. 

This PR adds the ability to pass an array of domains to `Clearance.configuration.cookie_domain`. When issuing a cookie, the server checks to see if the domain in the URI matches any of the configured domains -- if so, it uses the matching domain -- if not, it uses the first domain specified.

NB: the name of the setting probably ought to be pluralized (`cookie_domains`), but we'd need to add an alias to avoid introducing breaking changes. Didn't want to bother with the search & replace until I got some feedback on the concept.

Behavior when a single domain and `nil` are passed to `cookie_domain` is unaltered. This PR should ease headaches when phasing out some bad architectural decisions without affecting people who are Doing It Right.

### DIFF
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -115,10 +115,12 @@ module Clearance
         value: remember_token
       }
 
-      allowed_domains = Array(Clearance.configuration.cookie_domain).map(&:downcase)
+      allowed_domains = Array(Clearance.configuration.cookie_domain).dup
+      allowed_domains.map!(&:downcase)
 
       if allowed_domains.any?
-        value[:domain] = allowed_domains.delete(extract_base_host(get_request_uri))
+        base_host = extract_base_host(get_request_uri)
+        value[:domain] = allowed_domains.delete(base_host)
         value[:domain] ||= allowed_domains.first
       end
 
@@ -129,12 +131,12 @@ module Clearance
 
     def extract_base_host(uri)
       hostname = URI.parse(uri).host
-      hostname.split('.').last(2).join('.').downcase
+      hostname.split(".").last(2).join(".").downcase
     rescue URI::InvalidURIError
     end
 
     def get_request_uri
-      @env['REQUEST_URI']
+      @env["REQUEST_URI"]
     end
   end
 end
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -115,11 +115,26 @@ module Clearance
         value: remember_token
       }
 
-      if Clearance.configuration.cookie_domain.present?
-        value[:domain] = Clearance.configuration.cookie_domain
+      allowed_domains = Array(Clearance.configuration.cookie_domain).map(&:downcase)
+
+      if allowed_domains.any?
+        value[:domain] = allowed_domains.delete(extract_base_host(get_request_uri))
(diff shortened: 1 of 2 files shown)


agraves:
Ping?

derekprior:
Hi Aaron,

Thanks for the PR. I haven't had time to look at this yet, as I've been
busy with client work lately. I may have some time on Friday, if not, it
will be the Friday after that.

agraves:
Cool thanks @derekprior no rush

lib/clearance/session.rb
```diff
@@ -115,11 +115,28 @@ def cookie_value
         value: remember_token
       }
 
-      if Clearance.configuration.cookie_domain.present?
-        value[:domain] = Clearance.configuration.cookie_domain
+      allowed_domains = Array(Clearance.configuration.cookie_domain).dup
```
derekprior: What do you think about moving the array wrapping to the reader in `Configuration`?

derekprior: or maybe the setter? I guess it depends on how else/where else we are using that configuration

lib/clearance/session.rb
```diff
@@ -115,11 +115,28 @@ def cookie_value
...
+      allowed_domains.map!(&:downcase)
+
+      if allowed_domains.any?
+        base_host = extract_base_host(get_request_uri)
+        value[:domain] = allowed_domains.delete(base_host)
+        value[:domain] ||= allowed_domains.first
```
derekprior: What is the scenario where the request domain isn't in our list of allowed domains and yet we still want to set the cookie to whatever the first domain is?

lib/clearance/session.rb
```diff
@@ -115,11 +115,28 @@ def cookie_value
...
       end
 
       value
     end
+
+    private
+
+    def extract_base_host(uri)
+      hostname = URI.parse(uri).host
+      hostname.split(".").last(2).join(".").downcase
```
derekprior: What about the need to set cookies based on subdomain? That is: `thoughtbot.myapp.com`?

derekprior:
I definitely appreciate the idea here, but I'm not sure we can move forward with this as is and I'm not sure how difficult moving forward with it would be...

For one, in the comments I pointed out that you may set your allowed domain down to the subdomain and the current code does not allow for this. But if you make the code consider the full host, you might miss the other case where you've got your allowed cookie domain set to `".example.com"`it should match `*.example.com` but I don't think it would with this code, correct?

What it feels like we would need is an object that, given a `cookie_domain` setting and a doamin determins if it matches, based on [the RFC](http://tools.ietf.org/html/rfc6265#section-5.1.3). I wonder if that exists already?

derekprior:
Going to close for now. @agraves ping me again if you'd like to move this forward.

PR #564, opened 2015-06-12
Simplify the check for forbidden access

mxie:
Instead of having two separate `before_filter`s that both render the same
failure notice, we'll just define a single callback method that checks on both
conditions. This clarifies what conditions warrant this particular failure
message.

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -3,8 +3,7 @@ require 'active_support/deprecation'
 class Clearance::PasswordsController < Clearance::BaseController
   skip_before_filter :require_login, only: [:create, :edit, :new, :update]
   skip_before_filter :authorize, only: [:create, :edit, :new, :update]
-  before_filter :forbid_missing_token, only: [:edit, :update]
-  before_filter :forbid_non_existent_user, only: [:edit, :update]
+  before_filter :ensure_existing_user, only: [:edit, :update]
 
   def create
     if user = find_user_for_create
@@ -76,6 +75,13 @@ class Clearance::PasswordsController < Clearance::BaseController
     find_user_by_id_and_confirmation_token
   end
 
+  def ensure_existing_user
+    unless find_user_by_id_and_confirmation_token
+      flash_failure_when_forbidden
+      render template: "passwords/new"
+    end
+  end
+
   def flash_failure_when_forbidden
     flash.now[:notice] = translate(:forbidden,
       scope: [:clearance, :controllers, :passwords],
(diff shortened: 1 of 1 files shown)


app/controllers/clearance/passwords_controller.rb
```diff
@@ -88,18 +87,19 @@ def flash_failure_after_update
       default: t('flashes.failure_after_update'))
   end
 
-  def forbid_missing_token
-    if params[:token].to_s.blank?
+  def check_for_forbidden_access
```
drapergeek: What do you think about `ensure_token_and_user`?

Currently, none of these actions are really forbidden access so the name feels a bit weird.

croaky:
This one takes me back! I think I introduced those filters in Feb 2009: https://github.com/thoughtbot/clearance/commit/80a270b129a3c7b4e6948393c8e957b2af39b588

Only benefit I can think of to have the separate methods is if someone wanted to override the private methods to do something different. It's possible some apps are doing that. Not technically part of the public API https://github.com/thoughtbot/clearance#controllers so should be okay if it breaks.

Looks good to me but @derekprior's more up to date.

app/controllers/clearance/passwords_controller.rb
croaky: Some history on the `forbid` / `forbidden` words: https://robots.thoughtbot.com/forbidden-kisses-http-fluency-in-clearance

Used to be a 403 when these things happened.

mxie:
@croaky Yeah, I definitely considered the fact that it'd break for anyone who was overriding those private methods. However, the fact that they're not part of the public API (as you said) and we're considering this as part of the 2.0 release, I think breakage/non-backwards-compatibility is acceptable.

app/controllers/clearance/passwords_controller.rb
derekprior: +1 to "ensure_token_and_user" or even just "ensure_user_present" or something.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -88,18 +87,19 @@ def flash_failure_after_update
...
     end
   end
 
-  def forbid_non_existent_user
-    unless find_user_by_id_and_confirmation_token
-      flash_failure_when_forbidden
-      render template: 'passwords/new'
-    end
+  def missing_token?
+    params[:token].to_s.blank?
```
derekprior: Any idea what we need `to_s` for here?

app/controllers/clearance/passwords_controller.rb
```diff
@@ -88,18 +87,19 @@ def flash_failure_after_update
       default: t('flashes.failure_after_update'))
   end
 
-  def forbid_missing_token
-    if params[:token].to_s.blank?
+  def check_for_forbidden_access
+    if missing_token? || user_does_not_exist?
```
derekprior: What do you think about something like:

```ruby
unless find_user_by_id_and_confirmation_token
  #...
end
```

And then moving the empty param check into `find_user_by_id_and_confirmation_token` itself? Would save us a couple of private methods and is more in line with what we want, I think - basically you shouldn't ever find a user here without them having a cofnirmation token set.

derekprior:
Good consolidation. Breaking changes are a fine line to walk when it comes to an engine. Almost everything could be considered a breaking change.

I cant think of a use case where you'd want to handle both of these seperately. I think I'm okay with the consolidation here and okay with doing it in 1.11.

app/controllers/clearance/passwords_controller.rb
mxie: That's a great point! I suspect that the token check was initially put in place to kind of short-circuit the checks or something.

mxie:
Thanks for the feedback, @drapergeek @croaky and @derekprior. I've pushed up some changes - ready for re-review.

croaky:
:+1:

derekprior:
I like it. :+1:

PR #565, opened 2015-06-12
Provide helpers to be used in view specs

derekprior:
This brings versions of the controller spec helpers `sign_in` and
`sign_in_as` to view specs and does so in a way that avoids people
getting [errors from verified partial doubles][1]. The downside of this
approach is that it changes the subject under test (the view) by
introducing these helper methods directly on it. I don't think that's
much different than controller helper methods, though.

There is potential for confusion among users that might expect calling
`sign_in` or `sign_in_as` would come with all of the same side effects
that calling these methods from a controller or controller spec would.
In reality, they are just fancy ways to set the `current_user` local
variable and have no side effects. I feel that the potential confusion
from this has less impact than having to remember different helper
method names depending on whether you are in a controller spec or a view
spec.

As part of this change, I moved the existing
`Clearance::Testing::Helpers` module to
`Clearance::Testing::ControllerHelpers` and deprecated the old name.
This is a more accurate name and allows us to isolate the `ViewHelpers`
which are not useful in test_unit.

[1]: https://github.com/rspec/rspec-rails/issues/1076

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,23 @@ sign_in_as(user)
 sign_out
 ```
 
+### View and Helper Spec Helpers
+
+Does the view or helper you're testing reference `signed_in?`, `signed_out?` or
+`current_user`? If you `require 'clearance/rspec'`, you will have the following
+helpers available in your view specs:
+
+```ruby
+sign_in
+sign_in_as(user)
+```
+
+These will make the clearance view helpers work as expected by signing in either
+a new instance of your user model (`sign_in`) or the object you pass to
+`sign_in_as`. If you do not call one of these sign in helpers or otherwise set
+`current_user` in your view specs, your view will behave as if there is no
+current user: `signed_in?` will be false and `signed_out?` will be true.
+
 ## Contributing
 
 Please see [CONTRIBUTING.md].
--- a/lib/clearance/testing/helpers.rb
+++ b/lib/clearance/testing/helpers.rb
@@ -1,31 +1,15 @@
+require "clerance/testing/controller_helpers"
+
 module Clearance
   module Testing
+    # @deprecated Use Clearance::Testing::ControllerHelpers
     module Helpers
-      def setup_controller_request_and_response
-        super
-        @request.env[:clearance] = Clearance::Session.new(@request.env)
-      end
-
-      def sign_in
-        unless defined?(FactoryGirl)
-          raise(
-            RuntimeError,
-            "Clearance's `sign_in` helper requires factory_girl"
-          )
-        end
-
-        factory = Clearance.configuration.user_model.to_s.underscore.to_sym
-        sign_in_as FactoryGirl.create(factory)
-      end
-
-      def sign_in_as(user)
-        @controller.sign_in user
-        user
-      end
-
-      def sign_out
-        @controller.sign_out
-      end
+      warn(
+        "#{Kernel.caller.first} [DEPRECATION] Clearance::Testing::Helpers is "\
+        "deprecated and has been replaced with " \
+        "Clearance::Testing::ControllerHelpers. Require " \
+        "clearance/testing/controller_helpers instead."
+      )
     end
   end
 end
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -363,6 +363,23 @@ sign_in_as(user)
 sign_out
 ```
 
+### View Spec Helpers
+
+Does the view your testing reference `signed_in?`, `signed_out?` or
+`current_user`? If you `require clearance/rspec`, you will have the following
```
calebthompson: `require "clearance/rspec"`

lib/clearance/testing/helpers.rb
```diff
@@ -1,31 +1,14 @@
...
-
-      def sign_in_as(user)
-        @controller.sign_in user
-        user
-      end
-
-      def sign_out
-        @controller.sign_out
-      end
+      warn(
```
gabebw: Does this warning get output when this file is required, or when this module is included? It looks like it's when this file gets required. Given that, perhaps add "require clearance/testing_controller_helpers rather than clearance/testing/helpers"?

gabebw: Or! Wrap this warning in `included do`

spec/clearance/testing/view_helpers_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(view.current_user).to eq user
+    end
+  end
+
+  def test_view_class
+    Class.new do
+      include Clearance::Testing::ViewHelpers
+
+      def view
+        @view ||= extend Clearance::Testing::ViewHelpers::CurrentUser
```
gabebw: Whoa, `extend` is an expression? What does it return?

spec/views/view_helpers_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "spec_helper"
+
+describe "Clearance RSpec view spec configuration" do
```
gabebw: Perhaps `type: :view` on this line would make it easier to figure out exactly what it's testing?

gabebw: Also, do you want a `type: :helper` spec to test that it's included in helper specs too?

gabebw:
I'm a big fan. If Clearance weren't stubbing the view SUT, I'd do it, so this is just pure convenience.

spec/clearance/testing/view_helpers_spec.rb
derekprior: The extended object.

gabebw: Oh, so this is extending the `Class.new`? So it's...returning the extended anonymous eigenclass of an anonymously defined class?

lib/clearance/testing/helpers.rb
derekprior: Hmmm. Interesting question.

This would get output when the file was required. There were never any instructions to require this file directly, as far as I can tell. The instructions previously were to require `clearance/rspec` or `clearance/test_unit` which required and included this file as necessary.

I don't expext users to see this deprecation. If they do, it's likely that they'll want to know both what to require instead AND what they should be `including` once they have done so.

spec/clearance/testing/view_helpers_spec.rb
derekprior: ... sure.

lib/clearance/testing/helpers.rb
gabebw: > I don't expext users to see this deprecation. If they do, it's likely that they'll want to know both what to require instead AND what they should be including once they have done so.

Yes, that was my thought too. Given that they likely won't see this, I don't know how fancy this message needs to be, but I think adding "require clearance/testing/controller_helpers instead" would help them.

spec/clearance/testing/view_helpers_spec.rb
gabebw: I guess this works then.

calleerlandsson:
I'm in favor of this change.

spec/views/view_helpers_spec.rb
derekprior: Good call.

spec/clearance/testing/view_helpers_spec.rb
gabebw: Eigenclass: http://madebydna.com/all/code/2011/06/24/eigenclasses-demystified.html

PR #566, opened 2015-06-12
Fix acceptance test errors under RSpec 3.3

derekprior:
When the clearance test suite is run with RSpec 3.3, we get
`uninitialized constant RSpec::Rails::FixtureSupport`. There were some
`FixtureSupport` changes made in RSpec 3.3, but they should not impact
our acceptance tests at all.

When looking at the backtrace, I found gems from Clearance's test group
present when it should solely be gems included in our generated test
app. I switched to executing the acceptance test commands against a
clean bundler environment and everything works again.


spec/acceptance/clearance_installation_spec.rb
```diff
@@ -69,7 +69,10 @@ def successfully(command, silent = true)
       silencer = ""
     end
 
-    return_value = system("#{command} #{silencer}")
+    return_value = Bundler.with_clean_env do
+      system("#{command} #{silencer}")
+    end
```
croaky: Makes sense.

croaky:
Looks good to me.

mxie:
Looks good to me too.

derekprior:
Not waiting for travis. He seems to be napping...

PR #570, opened 2015-06-26
Only use table transformations in users migration

calleerlandsson:
Previously, the create users migration template was using [table
transformations] for adding columns and [schema statements] for adding
indices.

This commit updates the migration to use table transformations for both of
these purposes.

[table transformations]: http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Table.html
[schema statements]: http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html


derekprior:
:+1:

PR #571, opened 2015-07-17
Move remember token test to model specs

mxie:
The controller spec shouldn't know anything about regenerating the remember
token, so the test should be moved over to the User model spec instead.


derekprior:
:+1:

PR #574, opened 2015-08-21
Document password strategies

derekprior:


lib/clearance/password_strategies/blowfish.rb
```diff
@@ -3,6 +3,7 @@
 
 module Clearance
   module PasswordStrategies
+    # @deprecated Use BCrypt or `clearance-deprecated_password_strategies` gem
```
gabebw: Maybe `Use {Clearance::PasswordStrategies::BCrypt}`, i.e. link to the actual module?

lib/clearance/password_strategies/sha1.rb
```diff
@@ -1,5 +1,6 @@
 module Clearance
   module PasswordStrategies
+    # @deprecated Use BCrypt or `clearance-deprecated_password_strategies` gem
```
gabebw: Same here - what about linking to the module?

lib/clearance/password_strategies/bcrypt_migration_from_sha1.rb
```diff
@@ -1,5 +1,6 @@
 module Clearance
   module PasswordStrategies
+    # @deprecated Use BCrypt or `clearance-deprecated_password_strategies` gem
```
gabebw: How about linking to the BCrypt module?

lib/clearance/password_strategies/bcrypt.rb
```diff
@@ -1,5 +1,11 @@
 module Clearance
   module PasswordStrategies
+    # Uses BCrypt to authenticate users and store encrypted passwords.
+    #
+    # The BCrypt cost, the measure of how many key expansion iterations BCrypt
```
gabebw: I prefer parentheses:

>  The BCrypt cost (the measure of how many key expansion iterations BCrypt will perform) is automatically set...

PR #575, opened 2015-08-21
Routes generator also disables internal routes

### LINKED ISSUE #539: Failed to generate clearance:views after clearance:routes
### How to reproduce
1. Create new app:
  ```
  rails _4.2.0_ new <app_name>
  cd <app_name>
  ```
    
2. Install *Clearance*:
  ```
  rails g clearance:install
  ```
    
3. Generate routes:
  ```
  rails g clearance:routes
  ```

4. Generate views:
  ```
  rails g clearance:views
  ```

The last instruction should reveal the issue.
I got the following lines of output:

```
/home/<user_name>/.rvm/gems/ruby-2.1.0/gems/actionpack-4.2.0/lib/action_dispatch/routing/route_set.rb:538:in `add_route': Invalid route name, already in use: 'sign_in' 
You may have defined two routes with the same name using the `:as` option, or you may be overriding a route already defined by a resource with the same naming. For the latter, you can restrict the routes created with `resources` as explained here: 
http://guides.rubyonrails.org/routing.html#restricting-the-routes-created (ArgumentError)
```

As a result, the `rails g clearance:views` command provides no changes to the project.

derekprior:
Prior to this change, if you ran the routes generator and then started
your app, it would fail to start because of duplicate named routes. You
were required to manually change your initializer to add `config.routes
= false`. Now the generator does this for you.

Closes #539

### DIFF
--- a/spec/app_templates/config/initializers/clearance.rb
+++ b/spec/app_templates/config/initializers/clearance.rb
@@ -1,0 +1,2 @@
+Clearance.configure do |config|
+end
--- a/lib/generators/clearance/routes/routes_generator.rb
+++ b/lib/generators/clearance/routes/routes_generator.rb
@@ -9,6 +9,14 @@ module Clearance
         route(clearance_routes)
       end
 
+      def disable_clearance_internal_routes
+        inject_into_file(
+          "config/initializers/clearance.rb",
+          "  config.routes = false\n",
+          after: "Clearance.configure do |config|\n",
+        )
+      end
+
       private
 
       def clearance_routes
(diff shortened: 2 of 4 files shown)


spec/app_templates/config/initializers/clearance.rb
```diff
@@ -0,0 +1,3 @@
+Clearance.configure do |config|
+  config.mailer_sender = "reply@example.com"
```
jferris: Is this related to the rest of the changes?

gabebw:
Hey great! :+1:

spec/app_templates/config/initializers/clearance.rb
derekprior: Yes, this is a template used by the generator specs. It's the pre-existing initializer for this test case.

jferris: Do we need the mailer sender? I found the extra information a little confusing.

derekprior: No. I suppose not.

spec/app_templates/config/initializers/clearance.rb
```diff
@@ -0,0 +1,2 @@
+Clearance.configure do |config|
```
derekprior: I want this here.

PR #577, opened 2015-08-28
Document Clearance testing helpers

derekprior:


seanpdoyle:
Could we provide a thorough commit message documenting the changes?

seanpdoyle:
Do these comments hook into HTML documentation generation?

mike-burns:
These are beautiful.

@seanpdoyle this is written using YARD syntax, which rubydoc.info makes use of.

lib/clearance/testing/deny_access_matcher.rb
```diff
@@ -1,10 +1,40 @@
...
   module Testing
+    # Provides matchers to be used in your controller specs.
+    # These are typically exposed to your controller specs by
+    # requiring `clearance/rspec` or `clearance/test_unit` as
+    # appropriate in your `rails_helper.rb` or `test_helper.rb`
+    # files.
     module Matchers
+      # The `deny_access` matcher is used to assert that a
+      #   request is denied access by clearance.
+      # @option opts [String] :flash The expected flash notice message. Defaults
```
seanpdoyle: Should we provide an example of what it would look like to use one of these options?

derekprior: I thought about that, but documenting aything beyond a contrived example would require a bit more code (including a sample clearance configuration) for a realistic example. In the end, I think the options are documented well enough for now.

PR #579, opened 2015-08-29
Document authorization module

derekprior:


lib/clearance/authorization.rb
```diff
@@ -65,14 +99,21 @@ def return_to
       end
     end
 
+    # @private
```
graysonwright: Is this RDoc syntax?

graysonwright:
Small question, but looks good to me!

lib/clearance/authorization.rb
derekprior: yardoc.

I'm pretty sure it's not going to generate what I want. Clearance is weird because some methods exposed to ruby as `private` are part of public API. But other methods that are technically public are private API.

Once I've got the documentation written, I'll circle back and make sure the metadata is correct for what I want generated. I just dont want that to get in the way of the documentation effort currently.

PR #582, opened 2015-09-02
Remove unnecessary ActiveSupport::Concern

derekprior:
This is a vanilla module. It does not need to be a concern.


halogenandtoast:
:+1:

PR #584, opened 2015-09-04
Document the Clearance::Engine class

derekprior:


BlakeWilliams:
LGTM

lib/clearance/engine.rb
```diff
@@ -2,6 +2,21 @@
 require "rails"
 
 module Clearance
+  # Makes Clearance behavior available to Rails apps on initialization. By using
+  # a Rails Engine rather than a Railtie, Clearance can automatically expose its
```
JoelQ: What does "automatically expose its routes" mean in this context? That we don't have to manually load them in the routes file?

derekprior: yes.

lib/clearance/engine.rb
```diff
@@ -2,6 +2,22 @@
...
+  # own routes and views to the hosting application.
+  #
+  # Requiring `clearance` (likely by having it in your `Gemfile`) will
+  # automatically require the engine. You can opt-out of Clearance's internal
+  # routes by using {Configuration#routes=}. You can override the Clearance
```
gylaz: What should `Configuration#routes=` be set to in the case of opting-out? `false`, `[]`, something else?

derekprior: `false`. In the rendered docs, this is a link to the documentation for that method, which says:

```ruby
    # Set to `false` to disable Clearance's built-in routes.
    # Defaults to `true`. When set to false, your app is responsible for all
    # routes. You can dump a copy of Clearance's default routes with
    # `rails generate clearance:routes`.
    # @return [Boolean]
    attr_writer :routes
```

PR #588, opened 2015-09-07
Document Clearance user module

derekprior:
Several areas of clearance, the user module being one, contain methods that
are marked as private in ruby but are actually part of the public API for
the gem. For instance the private methods in the User module are private so
that they aren't exposed publicly in the host application's user model, but
they are certainly available for use to developers.

For this reason, this change include changes to the yard options to allow
documentation of private methods. If a method in Clearance is truly
internal API then it should be marked with `@api private`. The yard options
will skip these.

Now that we're allowing yard to document private methods, we need to be
explicit about what we do not wish to be public api by tagging it. This
definitely increases the noise in documentation of private parts of library
code, but it's the only way I could find to allow us to document the methods
that have private visiblity but are actually public API.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,3 +1,4 @@
+--protected
 --private
 --hide-api private
 --exclude templates
--- a/.yardopts
+++ b/.yardopts
@@ -1,3 +1,5 @@
+--private
+--hide-api private
 --exclude templates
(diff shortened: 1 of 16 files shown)


JoelQ:
LGTM :+1: Nice work!

PR #592, opened 2015-09-22
Correct typo in example

pedrosmmoreira:
The example for `Authentication#sign_in` contains a small typo,
indicating that Status object yielded to the method's block responds to
`success` when in fact it responds to the predicate method `success?`.


derekprior:
Merged in 7a63a2a.

derekprior:
Thank you!

PR #594, opened 2015-10-09
used let syntax to declare variable(user) to remove duplicity in password controller spec

kamal-github:


derekprior:
Thanks for the PR. 

I'd prefer to have the setup inline. This saves typing at the cost of making the specs a tiny bit harder to read as a whole. 

This is defined in our style guide here: https://github.com/thoughtbot/guides/tree/master/best-practices#testing

PR #596, opened 2015-10-17
Adding translation for pt-BR (Brazilian Portuguese)

ramirovjr:
With this commit, users of the project will be able to use
translated messages for the Brazilian Portuguese language.

As this is a translation addon, no tests were added.

### DIFF
--- a/config/locales/clearance.pt-BR.yml
+++ b/config/locales/clearance.pt-BR.yml
@@ -1,0 +1,54 @@
+---
+pt-BR:
+  clearance_mailer:
+    change_password:
+      closing: Se você não fez essa solicitação, ignore este email. Sua senha
+        não foi alterada.
+      link_text: Alterar minha senha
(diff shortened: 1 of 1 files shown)


derekprior:
Thanks for this @ramirovjr. For the time being we've decided not to accept localization PRs into clearance itself. There are a couple of different reasons for this:

1. We're not really suited to check if changes to functionality would impact translated text.
2. If we add new keys for a feature, do we have to wait for all translations in order to ship?

We have [clearance-i18n](https://github.com/thoughtbot/clearance-i18n) which is both a gem that users can opt into and also meant as a starting point for people to extract the translations they may require. We haven't publicized it much because, to be honest, I'm not 100% satisfied with this approach either (versioning compatibility has me worried).

For the time being, at least, we will continue this approach. The PT-BR translations do already exist in clearance-i18n. You can open a PR there if there are differences (particularly if the existing ones are incorrect).

Thanks!

Issue #562, opened 2015-05-29
Create flash message for require_login?


samjbmason:
When a user tries to access a page which requires authentication, it would be great to have a flash message after it redirects them to the `sign_in_url` where we could say that a user needs to be logged in to perform that action.

derekprior:
Is being redirected the login form not a sufficient enough notification that the users needs to log in?

rodrigodealer:
I think it's a good idea. I'll make a PR and then you guys can evaluate it.

derekprior:
Closed with #597

PR #597, opened 2015-10-17
Flash message when require login after access denied

rodrigodealer:
I added a flash message in require_login method  (Authorization class). 
This message is displayed when a non-authenticated user is accessing a resource that needs an authenticated user.
I think it's helpful to add a message of what the user needs to do to access the resource.
This solves #562.

### DIFF
--- a/config/locales/clearance.en.yml
+++ b/config/locales/clearance.en.yml
@@ -12,6 +12,7 @@ en:
     failure_after_update: Password can't be blank.
     failure_when_forbidden: Please double check the URL or try submitting
       the form again.
+    failure_when_access_denied: Please, you need to login.
   helpers:
     label:
       password:
--- a/config/locales/clearance.en.yml
+++ b/config/locales/clearance.en.yml
@@ -12,7 +12,7 @@ en:
     failure_after_update: Password can't be blank.
     failure_when_forbidden: Please double check the URL or try submitting
       the form again.
-    failure_when_access_denied: Please, you need to login.
+    failure_when_not_signed_in: Please sign in to continue
   helpers:
     label:
       password:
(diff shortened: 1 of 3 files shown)


spec/controllers/permissions_controller_spec.rb
```diff
@@ -54,6 +54,12 @@ def show
 
       expect(subject).to deny_access(redirect: sign_in_url)
     end
+
+    it 'denies access to show and display a flash message' do
```
rodrigodealer: I was following the coding style in all other tests.

lib/clearance/authorization.rb
```diff
@@ -19,7 +19,7 @@ module Authorization
     #     end
     def require_login
       unless signed_in?
-        deny_access
+        deny_access(I18n.t('flashes.failure_when_access_denied'))
```
derekprior: In this case, we're specifically denying access because there is no signed in user. I don't think "failure_when_access_denied" is an appropriate key here because you can actually call `deny_access` anywhere for any reason. It's a public method.

What about something like, "failure_when_not_signed_in" or something?

spec/controllers/permissions_controller_spec.rb
derekprior: We're updating the style as we go forward. Please make the changes requested by hound.

config/locales/clearance.en.yml
```diff
@@ -12,6 +12,7 @@ en:
     failure_after_update: Password can't be blank.
     failure_when_forbidden: Please double check the URL or try submitting
       the form again.
+    failure_when_access_denied: Please, you need to login.
```
derekprior: How about, "Please sign in to continue"

derekprior:
Good idea. I had some comments, if you want to take a crack at updating so I can merge. Thanks!

rodrigodealer:
@derekprior Thank you for your answer. I added some commits. Please, review it. :)

derekprior:
Merged in 7e7936d4e689fb45db22e56bb39a7b7ca475f294. Thank you so much!

PR #598, opened 2015-10-17
Properly indent routes

duleorlovic:
Nice tool, but when I use generators daily, I have to properly indent routes so my clients do not blame me :[


derekprior:
Thanks for bringing this to my attention. I have a PR that fixes this in a different way. See #603, which I will merge soon.

duleorlovic:
OK. Just to note that there are other places in this pull request that
needs proper indent.

derekprior:
Ah, I missed that. Fixing separately. Thanks for reminding me!

PR #603, opened 2015-11-05
Clean up indentation of generated routes

derekprior:
The `route` generator action provided by rails assumes a single line
route injection. The first line is automatically indented, but the
following lines are not.

I thought about injecting the routes one line at a time, but that
results in weird output from the generator because we format some long
lines in the routes to avoid going over 80 characters.

I decided to override the `route` action provided by default rails
generators to do exactly what we need. This also has the benefit of
cleaning up the output of the generator as well.

### DIFF
--- a/lib/generators/clearance/routes/templates/routes.rb
+++ b/lib/generators/clearance/routes/templates/routes.rb
@@ -1,12 +1,12 @@
-resources :passwords, controller: "clearance/passwords", only: [:create, :new]
-resource :session, controller: "clearance/sessions", only: [:create]
+  resources :passwords, controller: "clearance/passwords", only: [:create, :new]
+  resource :session, controller: "clearance/sessions", only: [:create]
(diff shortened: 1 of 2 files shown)


lib/generators/clearance/routes/templates/routes.rb
```diff
@@ -1,12 +1,12 @@
...
+  resource :session, controller: "clearance/sessions", only: [:create]
 
-resources :users, controller: "clearance/users", only: [:create] do
-  resource :password,
-    controller: "clearance/passwords",
-    only: [:create, :edit, :update]
-end
+  resources :users, controller: "clearance/users", only: [:create] do
+    resource :password,
+      controller: "clearance/passwords",
```
derekprior: I prefer the way this reads in our routes file. Recent changes to our guides make this rule obsolete.

PR #606, opened 2015-11-13
Fix mailer delivery expectation

imtayadeway:
3ca7a6b20a73f90b68fbfdb5a794d6caa1d941ef introduced a change in
`visitor_resets_password_spec.rb.tt` to break up the expectation on the
email's body into an `html_part` and `text_part`. The block that finds
the email however was only considering the match to the `text_part`
because it was not chained to the other conditions, so the subject and
html part were effectively ignored.

This commit chains all the conditions together to ensure they are
finding the right thing.

***

Note: I couldn't find anywhere where this code was being executed in the test suite, but I did verify that the change passed in my app, and that inverting the conditions of the first expression (i.e. by using `!~` instead of `=~`s) had no effect before, but did fail the test after. Hope that suffices!

### DIFF
--- a/lib/generators/clearance/specs/templates/features/clearance/visitor_resets_password_spec.rb.tt
+++ b/lib/generators/clearance/specs/templates/features/clearance/visitor_resets_password_spec.rb.tt
@@ -47,7 +47,7 @@ feature "Visitor resets password" do
     message = ActionMailer::Base.deliveries.any? do |email|
       email.to == [recipient] &&
         email.subject =~ /#{subject}/i &&
-        email.html_part.body =~ /#{body}/
+        email.html_part.body =~ /#{body}/ &&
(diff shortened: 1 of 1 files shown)


derekprior:
One of the tests that gets run on a full rake generates a new app, installs clearance and the generated specs, and runs the tests. So that's how this gets tested. Thanks for the patch!

imtayadeway:
@derekprior strange, I couldn't get it to fail when running the rake task, though I did have some pains in getting myself set up (particularly because I vendorize my gems by default - addressed in appraisal here - https://github.com/thoughtbot/appraisal/commit/91eb5d63fdbc7237c6e1715b5569ac85a727a5d6 - and because my `railsrc` was interfering. I'll see if I can address some of those issues in a follow up PR(s)). Thanks!

PR #607, opened 2015-11-18
Ignore the railsrc if present

imtayadeway:
A railsrc may make unwanted configurations when generating the test app,
so this change adds the `--no-rc` flag which will ignore it.


derekprior:
merged in 9c7d85e42d38f32f399b2503ee6549eceb30dcdb, thank you!

PR #609, opened 2015-11-23
add translation value used by the reset password mailer

bwanicur:
The documentation suggests that you can alter the email subject of the reset password mailer inside the config/locales/clearance.en.yml file, however, there is no default value there.  It would probably be more user friendly if the default value existed inside the config/locales/clearance.en.yml file.


derekprior:
Wow, that's a crazy non-conventional key we're using there.

Nice idea, I'll get this merged.

derekprior:
Merged in 463f468. Thanks so much for the contribution!

PR #613, opened 2015-12-05
if sign up is not allowed - redirect away from sign up form page. 

cjfinc:
and the same for create controller.

### DIFF
--- a/app/controllers/clearance/users_controller.rb
+++ b/app/controllers/clearance/users_controller.rb
@@ -4,19 +4,30 @@ class Clearance::UsersController < Clearance::BaseController
   skip_before_filter :authorize, only: [:create, :new]
 
   def new
-    @user = user_from_params
-    render template: "users/new"
+    if Clearance.configuration.allow_sign_up?
+      @user = user_from_params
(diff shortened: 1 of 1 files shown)


derekprior:
Thanks for the contribution.

If you have `allow_sign_up?` set to `false` then Clearance disables these routes. If you''ve run the `routes` generator to copy Clearance's routes locally then you have the power to redirect these (or delete these) from your routes file already.

Is there another use case we're not considering?

cjfinc:
Hi.
Yeah, I have a copy of routes.

My scenario are following:
As a site owner I want to on/off registration on site
so that I have an ability to open registration for limited time

And config option "config.allow_sign_up" is pretty handy for this.
But for local routes copy it requires my fix.

derekprior:
I've elected not to expand the use of `allow_sign_up?` at this time. You should be able to implement this in your own application in any number of ways - either by overriding the controller actions (calling `super` after your check) or perhaps with custom routing constraints and the like.

Again, thanks for the contribution it's just not something I'm comfortable incorporating at this point.

Issue #616, opened 2015-12-10
Duplicate cookie sent when using cookie domains


coryodaniel:
When using cookie domains two cookies were being sent to the browser:
1. a domain cookie
2. a regular cookie

Also, logging out would only the regular cookie to get deleted because the [sign out method](https://github.com/coryodaniel/clearance/blob/master/lib/clearance/session.rb#L80) ignored if the domain was set. This would cause one cookie to still be set in the browser after log out, and create bizarre bouncy redirects when you logged in again (because the remember token was set the the previous session's remember token).

Here is my config:

``` ruby
Clearance.configure do |config|
  config.cookie_name = "remember_token_#{Rails.env}"
  config.mailer_sender = %("Oh hi" <hello@example.com>)
  config.password_strategy = Clearance::PasswordStrategies::BCryptMigrationFromSHA1
  config.cookie_domain = ".example.com"
  config.cookie_expiration = ->(cookies) { cookies['remember_me'] ? 1.year.from_now.utc : nil }
  config.secure_cookie = true
  config.httponly = true
end

```

Note: this is specifically for version 1.8.x, I wasn't able to upgrade to 1.12 due to a gem conflict, but the sections of changed code are the same.

I created a [branch](https://github.com/coryodaniel/clearance/tree/wonky_cookie_domain) with the fixes. It broke some expectations and I didnt put coverage in (because I'm a huge jerk). I wanted to get this ticket opened as a reminder while I send this up to my staging environment to see if its working across all my multi tenant domains.

I'll do a proper set of coverage and a PR against master once I have confirmed. Which branch should I make a PR against for the 1.8.x versions? Are you guys working on a 1.8.2?

<img width="881" alt="results from session create -clearing cookies and then logging into" src="https://cloud.githubusercontent.com/assets/11587/11706164/0d51a566-9eab-11e5-8993-1ed91c78a402.png">

coryodaniel:
For the failing tests, a lot of the tests are written around a method #add_cookie_to_headers which is called in the middleware Clearance::RackSession. 

This isn't actually required as rails writes the Set-Cookie header for anything in the cookies collection and is what causes the double remember token cookie being sent in the event a domain cookie is used:

[...]

PR #617, opened 2015-12-10
Fix for duplicate cookie when using cookie domain option

coryodaniel:
See this issue: https://github.com/thoughtbot/clearance/issues/616

When using cookie domains two cookies were being sent to the browser:
1. a domain cookie
2. a regular cookie

Also, logging out would only the regular cookie to get deleted because the sign out method ignored if the domain was set. This would cause one cookie to still be set in the browser after log out, and create bizarre bouncy redirects when you logged in again (because the remember token was set the the previous session's remember token).


derekprior:
@coryodaniel: Thanks for looking into this. These seems like really large changes and I have to admit I don't fully understand the problem right now. Is there any way you can either:

1. submit a failing test case, with no code changes, that describes the problem.
2. provide a link to a minimal reproduction application that I can play with.

When I have the time to do so, I'll probably try the 2nd one myself to better understand this, but that likely won't be for some time.

coryodaniel:
@derekprior Hey I should be able to jump on this in a weekend or two. You can take a vanilla generated rails app and use the current version of clearance or 1.8.1 and the issue will occur. Its pretty much just turning on cookie domains. After that you'll see the cookie sent twice, given you are on 'www.example.com' you'll get a remember token for "www.example.com" and "example.com"

coryodaniel:
Forgot all about this. Was it merged?

MSchmidt:
Seems like this was closed but not merged? Not sure. Maybe @derekprior knows. Clearance 1.15.1 has same issue though.

derekprior:
This is not merged. The best way to move forward would be a single failing test case or minimal reproduction sample app. At that point, I'll better understand the issue and hopefully be able to provide some guidance on what I think the best change would be in a way that limits the scope as much as possible.

Thanks.

MSchmidt:
I've posted a reproduction sample app to #616. Hope this helps.

PR #618, opened 2015-12-11
Make generated code follow the current style guide

teoljungberg:
Since these migrations are generated, we want them to follow the current
style guide. So when opening a PR, there will be no violations reported
from [Hound].

[Hound]: https://houndci.com


derekprior:
Thanks @teoljungberg

PR #623, opened 2016-01-06
Start testing against Rails 5 beta 1

derekprior:
We know it's going to fail, but I'd like to have Travis start running it
for us so we can monitor progress as we make changes.

This also changes the Travis build so it does not use appraisal
directly, but rather uses the gemfiles appraisal generates in the Travis
matrix. This will make it clear what builds are in error.


jferris:
> This will make it clear what builds are in error.

It should also make the builds run in parallel!

PR #624, opened 2016-01-07
Make create_users migration Rails 3.2 compatible

derekprior:
This reverts commit d080a5b2f1adff8fcb104e39b1b192b7c901d689. The
`index` method was introduced to `ActiveRecord::TableDefinition` in
Rails 4.0, but Clearance 1.x supports back to 3.2.

This should have been a test failure, but it seems our tests are not
executing in the environment we think they are. I'm working through that
issue in some changes to the test suite in PR #623.


derekprior:
Correction - the gemspec says we support 3.1 as well (but we do not test against it, thus my mistake).

Issue #621, opened 2015-12-24
Rails 5 raises errors on skip_before_filter if the callback hasn't been made


j-dexx:
[Method link] (https://github.com/rails/rails/blob/master/activesupport/lib/active_support/callbacks.rb#L632) for rails where it occurs

This means if `before_action :require_login` hasn't been set in the application controller clearance's sessions, passwords and users controller will raise errors.

Can be remedied by either updating the documentation to say that you need to add them in, or by adding them to the authorization module.

    module Clearance
      module Authorization
        extend ActiveSupport::Concern

        included do
          # hide_action :authorize, :deny_access, :require_login
          before_action :authorize, :require_login      
        end
      end
    end

Hide action commented out as support for this was dropped in rails 5.

derekprior:
It seems like we should be able to pass `raise: false` as an option to `skip_before_filter`.

j-dexx:
@derekprior Yes, that works nicely and is also fine with rails 4.

derekprior:
Closed by 23d93b8

PR #626, opened 2016-01-10
Add raise: false to skip_before_filter

j-dexx:
This fixes issue #621 

In rails 5 if the filter has not been defined then
an error will be raised.

This change addresses the need by:

Adding raise: false to the passwords, sessions and
users controller


derekprior:
Merged in 23d93b8 - Thanks!

PR #627, opened 2016-01-11
Remove use of `hide_action`

derekprior:
We were previously relying on `hide_action` to hide clearance helpers,
which are public in their own modules, from appearing as routable
actions in the controllers they are mixed into.

`hide_action` has been removed from Rails. We can use `private` instead.

### DIFF
--- a/lib/clearance/authentication.rb
+++ b/lib/clearance/authentication.rb
@@ -4,7 +4,7 @@ module Clearance
 
     included do
       helper_method :current_user, :signed_in?, :signed_out?
-      hide_action(
+      private(
         :authenticate,
         :current_user,
         :current_user=,
--- a/lib/clearance/authorization.rb
+++ b/lib/clearance/authorization.rb
@@ -3,7 +3,7 @@ module Clearance
     extend ActiveSupport::Concern
 
     included do
-      hide_action :authorize, :deny_access, :require_login
+      private :authorize, :deny_access, :require_login
     end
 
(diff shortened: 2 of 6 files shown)


jferris:
The `hide_action` to `private` conversion makes sense to me.

It looks like some other changes were included in this pull request, and I can't figure out how they're related to `hide_action` being removed. Should those changes be in a separate commit?

derekprior:
They're related because our test helpers made use of the methods being public. e.g. `@controller.sign_in`. 

This raises an interesting point: users could have been relying on the same in their own apps. However, I think of this as a bug we're fixing more than a feature that we're removing. I have doubts people are doing this in their own apps and think we can proceed with this as a 1.x change.

jferris:
Okay, that makes sense.

I think I agree with you about removing without deprecating, but it is a little sticky. Those controller methods are included in the [public documentation](http://www.rubydoc.info/github/thoughtbot/clearance/Clearance/Controller) with no indication that they aren't intended to be called publicly. It seems like an oversight that we didn't make them private to begin with, but....we didn't.

It would be a huge pain to deprecate them, though, since the methods aren't deprecated. Just calling them with an explicit receiver is.

derekprior:
I think most people expect controller helpers to be "private". I can't imaging an application use case where this would be a problem. I think it's more likely to be a problem in tests, but even then, I suspect updating our own helpers is sufficient.

The Rails 5 changes are going to be numerous and will likely require a beta release. So we can test the waters there a bit.

PR #628, opened 2016-01-11
Add spanish i18n

alejandropereira:


derekprior:
We had previously split this out to https://github.com/thoughtbot/clearance-i18n/, but in hindisght I'm not entirely sure that's a great solution. I might want to just merge these in here, though it means there will be some additional overhead or perhaps untranslated values when things change...

derekprior:
@alejandropereira What's your experience been with i18n for other engines (if any)? Do they provide translations in the library?

jferris:
> We had previously split this out to https://github.com/thoughtbot/clearance-i18n/, but in hindisght I'm not entirely sure that's a great solution.

I don't feel strongly either way, but one of the reasons we created that repo was so that we could have separate permissions. The idea was to give translators push access to just the translations.

JulienItard:
Any news about it ? I added clearance-i18n on rubygem. Is it useless or ?
IMO, we should keep translations files inside main repo, here.

Julien

derekprior:
For now, let's keep them in `clearance-i18n`.

The reasons for splitting them out originally were:

1. decouple i18n changes from clearance releases
2. allow people who know these languages to have commit on the i18n repo themselves (because I can't review Spanish).

For Clearance 2.0, I'd like to explore the possibility of using the `clearance-i18n` repo as a git submodule directly in this repo.

I'm not sure about how well that'd work or the security of it, though.

derekprior:
@JulienItard Do you want to merge this into clearance-i18n for the time being?

PR #630, opened 2016-01-13
remove duplicacy in comment in token

kamal-github:


lib/clearance/token.rb
```diff
@@ -1,5 +1,5 @@
 module Clearance
-  # Random token used for password reset tokens and password reset tokens.
+  # Random token used for password reset tokens.
```
derekprior: This should actuall be "and remember tokens". Can you make that change?

derekprior:
I went ahead and made the suggested change. This has been merged in fbf026e. Thanks for the contribution!

kamal-github:
Thanks! @derekprior

PR #631, opened 2016-01-18
update ReadMe Docs

estelleccl:
## Getting Started

Clearance is a Rails engine tested against Rails `>= 3.2` and Ruby `>= 1.9.3`.

You can add it to your Gemfile with:

```sh
gem 'clearance'
```

Run the bundle command to install it.

After you install Clearance, you need to run the generator:

```sh
$ rails generate clearance:install
```

The generator will:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,17 +16,25 @@ monitored by contributors.
 [GitHub Issues]: https://github.com/thoughtbot/clearance/issues
 [Stack Overflow]: http://stackoverflow.com/questions/tagged/clearance
 
-## Install
+## Getting Started
 
 Clearance is a Rails engine tested against Rails `>= 3.2` and Ruby `>= 1.9.3`.
-To get started, add Clearance to your `Gemfile`, `bundle install`, and run the
-`install generator`:
+
+You can add it to your Gemfile with:
+
+```sh
+gem 'clearance'
+```
+
+Run the bundle command to install it.
+
+After you install Clearance, you need to run the generator:
 
 ```sh
 $ rails generate clearance:install
 ```
 
-The generator:
+The generator will:
 
 * Inserts `Clearance::User` into your `User` model
 * Inserts `Clearance::Controller` into your `ApplicationController`
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ $ rails generate clearance:install
 
 The generator will:
 
-* Inserts `Clearance::User` into your `User` model
-* Inserts `Clearance::Controller` into your `ApplicationController`
+* Insert `Clearance::User` into your `User` model
+* Insert `Clearance::Controller` into your `ApplicationController`
 * Creates an initializer to allow further configuration.
 * Creates a migration that either creates a users table or adds any necessary
   columns to the existing table.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,17 +16,25 @@ monitored by contributors.
...
+After you install Clearance, you need to run the generator:
 
 ```sh
 $ rails generate clearance:install
 ```
 
-The generator:
+The generator will:
 
 * Inserts `Clearance::User` into your `User` model
```
magikid: Could you change this line and the next to Insert (instead of Insert**s**)?

estelleccl: @magikid thanks. :) Have made the changes.

README.md
```diff
@@ -16,22 +16,30 @@ monitored by contributors.
...
-The generator:
+The generator will:
 
-* Inserts `Clearance::User` into your `User` model
-* Inserts `Clearance::Controller` into your `ApplicationController`
-* Creates an initializer to allow further configuration.
-* Creates a migration that either creates a users table or adds any necessary
+* Insert `Clearance::User` into your `User` model
+* Insert `Clearance::Controller` into your `ApplicationController`
+* Creates an initializer file to allow further configuration.
```
magikid: Sorry to seem pedantic but could you please also change Creates to Create on this line and the next?

estelleccl: okie done :)

magikid:
:+1:  I like having the commands there in the README so that I can copy/paste them with ease.  What do you think @derekprior?

README.md
```diff
@@ -16,22 +16,30 @@ monitored by contributors.
...
+The generator will:
 
-* Inserts `Clearance::User` into your `User` model
-* Inserts `Clearance::Controller` into your `ApplicationController`
-* Creates an initializer to allow further configuration.
-* Creates a migration that either creates a users table or adds any necessary
+* Insert `Clearance::User` into your `User` model
+* Insert `Clearance::Controller` into your `ApplicationController`
+* Create an initializer file to allow further configuration.
+* Create a migration file that either create a users table or adds any necessary
```
derekprior: the second `create` here should be `creates`.

Also, I'm not sure we need the word "file" here. The generator output itself makes it pretty clear these are files, I think?

derekprior:
To me, the updated tense in the list items looks at first glance like a list of things I'm supposed to do. The previous tense made it clear that something else is doing it for me. At least that's my ipression. @estelleccl, what was your intention with that part of the change?

estelleccl:
Hi @derekprior, I'm a web developer bootcamp mentor. Majority of my students are new developers and clearance is the first gem they will be using.

I know it seem like a list of things they supposed to do but as for them everything is new. 

So I was wondering, having the commands there in the README so that they can understand the steps with ease and also good for beginner as well.

derekprior:
What I meant was - that list (the actuall bulleted list in this PR) could easily be scanned as a list of things *I* need to do to use this gem, when in reality it's a list of things the gem does for me when I run the install generator.

The rest of it I like. I'm just nitpicking of the changes from `Installs` to `Install`

estelleccl:
would changing `The generator will:` to `The command above will` help? 

I think as long as we don't highlight it as a command like ```gem 'clearance'```, it wouldn't look like a list of things I need to do.

estelleccl:
you are right, why dont we change it to become a paragraph instead.

derekprior:
I liked the bulleted list version, so I went back to that, but the other changes have been merged in a38ef7d. Thank you for the contribution!

PR #632, opened 2016-01-18
Don't rely on RSpec monkey patching for feature method

mikeastock:
The default spec_helper's suggested config sets `disable_monkey_patching!`.
This causes specs generated by Clearance to fail out of the box. This change
should be backwards compatible with older versions of RSpec as well.

[This](https://github.com/rspec/rspec-core/pull/1647) PR added the
`disable_monkey_patching!` config as a suggested default.


derekprior:
Thanks!

PR #639, opened 2016-02-03
Clean up Rails 5 deprecations

derekprior:
I believe these are all of the runtime deprecations we're getting under Rails
5. It's a bit hard to tell because there's also a lot of noise in the tests
from the decision to deprecate calling the HTTP verb methods in controller and
integration tests with positional arguments. That's "only" a deprecation we get
in our tests so we'll handle that separately.

Each of these fixes will remain a discrete commit with its own description. I'm
not thrilled with the `before_action` changes but this is going to be the last
minor version that supports Rails 3.x, so we can clean that up after release.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -20,6 +20,7 @@ end
 if RUBY_VERSION >= "2.2.0"
   appraise "rails50" do
     gem "rails", "~> 5.0.0.beta1"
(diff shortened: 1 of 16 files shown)


jferris:
This looks good to me. The `before_action` stuff is unfortunate, but I can't think of anything better.

PR #640, opened 2016-02-06
Fixed #635 - Fixed error when trying to create a user when `user_mode…

### LINKED ISSUE #635: If you configure user model, you cannot save the record on signup.
https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/users_controller.rb#L52

If you configure the user model in an initializer, records don't save. This is because the user params method is looking for `:user` instead of a symbol based on the configured user model. In my case `Person`.

jrguenther:
Fixed #635 - Fixed error when trying to create a user when `user_model` has been customized.

Appraisal passes for everything but rails50, but I don't those failures are caused by my code.

### DIFF
--- a/app/controllers/clearance/users_controller.rb
+++ b/app/controllers/clearance/users_controller.rb
@@ -50,6 +50,7 @@ class Clearance::UsersController < Clearance::BaseController
   end
 
   def user_params
-    params[:user] || Hash.new
+    user_model_sym = Clearance.configuration.user_model.name.parameterize.to_sym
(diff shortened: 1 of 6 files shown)


derekprior:
@jrguenther I took a slightly different approach by introducing a method on configuration for this. I'm happy with the testing provided by testing that method and I dont think we need a more end-to-end test. I left you as the commit author. Can you check out https://github.com/thoughtbot/clearance/pull/643 and let me know what you think? If it does the job we can get that merged soon.

PR #643, opened 2016-02-12
Use proper param name for overridden user model

derekprior:
The users controller was hard-coded to grab parameters from
`params[:user]`, but if you have changed your user model the key will be
something other than `:user`. We now get this key from configuration,
which derives it from the model name.

### DIFF
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -121,12 +121,20 @@ module Clearance
       end
     end
 
+    # The name of user parameter for the configured user model.
+    # This is derived from the `model_name` of the `user_model` setting.
+    # In the default configuration, this is `user`.
+    # @return [Symbol]
+    def user_parameter
+      user_model.model_name.singular.to_sym
+    end
+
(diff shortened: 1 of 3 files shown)


derekprior:
Fixes #635

lib/clearance/configuration.rb
```diff
@@ -121,12 +121,20 @@ def  user_actions
       end
     end
 
+    # The name of user parameter for the configured user model.
+    # This is derived from the `model_name` of the `user_model` setting.
+    # In the default configuration, this is `user`.
+    # @return [Symbol]
+    def user_parameter
+      user_model.model_name.singular.to_sym
```
jferris: This requires the user model to be an ActiveModel (or have `model_name`). I think that's reasonable, but is it already a requirement, or would this break backwards compatibility?

derekprior: It was a pre-existing requirement (see `user_id_parameter`) and I've said many times this gem has a dependency on ActiveRecord and other ORMs are not supported. So I think we're good.

PR #644, opened 2016-02-26
Test against Rails 5 master

derekprior:
Master includes a fix for https://github.com/rails/rails/issues/23645,
which is not yet included in a beta release.

When updating to master, I also had to explicitly set the ActiveJob
queue adapter to `inline` in tests, as it now defaults to `async`.

With these changes, the non-acceptance test suite is now green and has
no deprecations. Next up - the acceptance tests.

### DIFF
--- a/gemfiles/rails50.gemfile
+++ b/gemfiles/rails50.gemfile
@@ -13,7 +13,7 @@ gem "shoulda-matchers", "~> 2.8"
 gem "sqlite3", "~> 1.3"
 gem "timecop", "~> 0.6"
 gem "pry", :require => false
-gem "rails", "~> 5.0.0.beta1"
+gem "rails", :git => "https://github.com/rails/rails"
 gem "rails-controller-testing"
 gem "rspec-support", :github => "rspec/rspec-support"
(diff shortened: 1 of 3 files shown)


gemfiles/rails50.gemfile
```diff
@@ -13,7 +13,7 @@ gem "shoulda-matchers", "~> 2.8"
 gem "sqlite3", "~> 1.3"
 gem "timecop", "~> 0.6"
 gem "pry", :require => false
-gem "rails", "~> 5.0.0.beta1"
+gem "rails", :git => "https://github.com/rails/rails"
```
kenyonj: new hash syntax?

calebthompson:
Never used appraisals - is the gemfile generated?

gemfiles/rails50.gemfile
derekprior: It's generated.

derekprior:
Appraisal manages the gemfiles, yes. We need them checked in so we can get separate travis builds for each of them.

calebthompson:
:+1:

PR #645, opened 2016-02-26
Use versioned migrations when possible

derekprior:
Rails 5 deprecates inheriting directly from `ActiveRecord::Migration` in
favor of inheriting from `ActiveRecord::Migration[5.0]` where `5.0` is
the `major.minor` version of Rails that the migration was originally
written to support.

If we detect we are using a version of rails that supports this
nomenclature then we pass the current `major.minor` version to use.

### DIFF
--- a/lib/generators/clearance/install/templates/db/migrate/create_users.rb
+++ b/lib/generators/clearance/install/templates/db/migrate/create_users.rb
@@ -1,4 +1,4 @@
-class CreateUsers < ActiveRecord::Migration
+class CreateUsers < ActiveRecord::Migration<%= migration_version %>
   def change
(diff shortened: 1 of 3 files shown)


lib/generators/clearance/install/templates/db/migrate/create_users.rb
```diff
@@ -1,4 +1,4 @@
-class CreateUsers < ActiveRecord::Migration
+class CreateUsers < ActiveRecord::Migration<%= migration_version %>
```
derekprior: I assume this is because despite the `.rb` extension, `.erb` is actually supported here. I'm fine with this.

PR #648, opened 2016-02-26
Always use inline adapter in feature specs

derekprior:
Rails 5 defaults the ActiveJob adapter to `async` which makes it
difficult to test the password reset feature in a manner that supports
Rails 4.2 and 5.0 (and older versions as well, of course).

The simplest fix, for now, is to use the inline adapter in our tests so
we can observe the side effects we care about.

### DIFF
--- a/lib/generators/clearance/specs/templates/features/clearance/visitor_resets_password_spec.rb.tt
+++ b/lib/generators/clearance/specs/templates/features/clearance/visitor_resets_password_spec.rb.tt
@@ -3,6 +3,15 @@ require "support/features/clearance_helpers"
 
 RSpec.feature "Visitor resets password" do
   before { ActionMailer::Base.deliveries.clear }
+<% if defined?(ActiveJob) -%>
+
+  around do |example|
+    original_adapter = ActiveJob::Base.queue_adapter
+    ActiveJob::Base.queue_adapter = :inline
+    example.run
+    ActiveJob::Base.queue_adapter = original_adapter
+  end
+<% end -%>
 
   scenario "by navigating to the page" do
     visit sign_in_path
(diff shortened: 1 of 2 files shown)


lib/generators/clearance/specs/templates/features/clearance/visitor_resets_password_spec.rb.tt
```diff
@@ -3,6 +3,15 @@ require "support/features/clearance_helpers"
 
 RSpec.feature "Visitor resets password" do
   before { ActionMailer::Base.deliveries.clear }
+<% if defined?(ActiveJob) -%>
+
+  around do |example|
+    original_adapter = ActiveJob::Base.queue_adapter
+    ActiveJob::Base.queue_adapter = :inline
+    example.run
+    ActiveJob::Base.queue_adapter = original_adapter
```
jferris: Is it worth using `ensure` here so that this always gets reset, even if `example.run` raises an exception?

derekprior: I can do that. Is there a scenario where an exception not automatically handled by the example itself wouldnt immediately cause the test suite to crash?

jferris: I actually don't know how this works. If the test raises an exception or fails, does that get caught within `example.run` or higher up the stack? I was assuming that `around` errors bubbled up the same way `before` errors did, and that it would be reasonable to expect `example.run` to raise an error here.

derekprior: I added `raise Exception` to one of the example and added logging of the queue adapter before we run the example and after we reset the adapter. Everything executes as expected. I don't think `ensure` is necessary here.

PR #650, opened 2016-02-28
Call `find_by` instead of `where` and `first

zamith:
Why:

* There is no need to call `where` if we are going to call `first` right
  after it.

This change addresses the need by:

* Using the more intent revealing `find_by`

### DIFF
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -131,7 +131,7 @@ module Clearance
 
     # @api private
     def user_from_remember_token(token)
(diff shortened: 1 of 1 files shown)


derekprior:
Good point. Unfortunately, `find_by` is not supported in Rails < 4. The current construct gives us the greatest compatibility. 

When we cut over to 2.0 we can do this but I need to get a Rails 5 release out first. Closing for now.

zamith:
Oh, I didn't know it was not supported in Rails 3. Sorry about that.

PR #656, opened 2016-03-11
Allow Clearance::BackDoor to take a block

halogenandtoast:
If `User#to_param` is overriden the `Clearance::BackDoor` will fail
because it only looks the user up by id. By passing a block, this
functionality can be replaced to be more flexible. For instance, with a
User model that looks like:

    class User < ActiveRecord::Base
      def to_param
        username
      end
    end

You can specific the following backdoor in
`config/environments/test.rb`:

    config.middleware.use Clearance::BackDoor do |username|
      Clearance.configuration.user_model.find_by(username: username)
    end

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,20 @@ Usage:
 visit root_path(as: user)
 ```
 
+Additionally, if `User#to_param` is overridden, you can pass a block in
+order to override the default behavior:
+
+```ruby
+# config/environments/test.rb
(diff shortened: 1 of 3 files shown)


derekprior:
Looks good!

derekprior:
Thanks. This was merged as 663b039894546df01f841dda10b23625d27a84c1

PR #658, opened 2016-03-20
Remove html in text version of change password

developingchris:
closing p tag present in text version of the change password email by
default.

I know most people change this quickly, but when sending tests to show what the default was with a copywriter. My email client alerted me that there was html in my text version, and thought that was spammy.


derekprior:
Thanks for this!

PR #660, opened 2016-04-03
Add Defer Sign In Password Check feature

thiagoa:
## Introduction

This feature was born out of the need to implement an **Account Lock** mechanism, which works similarly to the one implemented in Devise (although much simpler).

Clearance's **Sign In** mechanism works with the following rules:
1. If a user exists in the database and the incoming password is **correct**,  a `user` object enters the `SignInGuard` stack.
2. If a user exists in the database and the incoming password is **incorrect**, a `nil` user object enters the `SignInGuard` stack.
3. If a user **does not exist in the database**, a `nil` user object enters the `SignInGuard` stack.
4. A user gets signed in _if_ the `current_user` object is not `nil` and the `SignInGuard` stack returns  `SuccessStatus`.

Unfortunately, the only way I was able to implement the **Account Lock** feature in my own app was by duplicating `SessionsController` and making an `AccountLockHandler` class, which acts as a pre-filter and runs the _account lock logic_ before handing out the user to `Clearance::Session#sign_in`, which in turns hands out the user to the `SignInGuard` stack.

The **Account Lock** feature requires working with an unauthenticated `user` object, so I could not use **Sign In Guards** because of the limitation described in item 2. I'm sure there are other features which would benefit from working with an unauthenticated user object.

My implementation felt particularly bad because:
- I had to duplicate Clearance's code in my app in order to get the feature working.
- Obeying the `current_user` rules described above, after my account lock pre-filter ran, felt a bit weird.
- This really feels like it should be handled by **Sign In Guards**.
## About the feature

I came up with this solution to make it possible to implement features such as **Account Lock** handlers _within_ **Sign In Guards**, while still keeping up with Clearance's philosophy of being small, simple and **extensible**, and not breaking up with existing Clearance contracts.

It works like this:
- The user enables a configuration option called `defer_sign_in_password_check`
- When signing in, the `Clearance::User.authenticate` method returns a `DeferredSignInUser` object. This is nothing but an augmented user wrapper (delegator), and does not check the user's password immediately, although it stores the password for a deferred check.
- The `SignInGuard` stack gets a `DeferredSignInUser` object – this allows guards to work with user objects regardless if they are authenticated or not.
- The `DeferredSignInUser#present?` method **only returns `true` if the user is present and authenticated**, so it works transparently with Clearance's current implementation.
- `ClearanceSession#signed_in?` delegates to `user#present?`, so `DefaultSignInGuard` will still do its job correctly without needing any changes.
## Alternative implementations

I built the feature in an opt-in manner, because I felt it would be a breaking change about how authentication and Sign In Guards work. In my opinion we should consider if this should indeed be the default behavior in future major versions, in order to extend the usefulness of Sign In Guards. That said, I'm OK with the way the feature got implemented in this PR.

There's an alternative, more explicit way to implement the feature, but it requires adding checks to `Clearance.configuration.defer_sign_in_password_check?` in lots of other spots, and ends up generating more code branches. I like the short and simple solution of having an augmented user, though a bit more implicit, but nevertheless well documented.
## Observations

If you have any questions, suggestions, comments, fixes, implementation ideas, anything, please don't hesitate to ask me. I've also left some descriptive text in the commits, to aid in answering possible questions.

Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Clearance.configure do |config|
   config.secure_cookie = false
   config.sign_in_guards = []
   config.user_model = User
+  config.defer_sign_in_password_check = false
 end
 ```
 
@@ -307,6 +308,39 @@ class EmailConfirmationGuard < Clearance::SignInGuard
   def unconfirmed?
     signed_in? && !current_user.confirmed_at
   end
+end
+```
+
+### Deferring Sign In Password Check
+
+You can also enable the `SignInGuard` stack to receive an unauthenticated _but_
+existing user object, thus deferring its password check to happen at the
+`SignInGuard` stack.  Clearance's default behavior is to represent an
+unauthenticated user as `nil`, but this is undesirable if you need to work with
+existing `user` objects regardless if the incoming password is correct or not.
+So this feature enables Sign In Guards to deal with concerns such as _locking
+an account after a certain number of failed attempts_.
+
+If you want to use this feature you'll need to enable it:
+
+```ruby
+Clearance.configure do |config|
+  config.defer_sign_in_password_check = true
+end
+```
+
+With this feature enabled the `Clearance::Session#current_user` method will
+hold a `DeferredSignInUser` object _at sign in time_, which can be used within
+your Sign In Guards as regular user objects.
+
+`DeferredSignInUser` wraps and delegates to your `user` object, and its
+`present?` method only returns `true` if the user is present **and** also
+authenticated. To check if there exists an underlying user to work with, you
+can use the `nil?` method within your custom Sign In Guards:
+
+```ruby
+if current_user.nil?
+  # No user found in the database
 end
 ```
 
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -85,6 +85,20 @@ module Clearance
     # @return [ActiveRecord::Base]
     attr_accessor :user_model
 
+    # Controls whether user password checks should be deferred to happen at the
+    # sign in guard stack. This feature enables a DeferredSignInUser object to
+    # enter within sign in guards, regardless if the user is authenticated or
+    # not.  Clearance's default behavior is to represent an unauthenticated
+    # user as nil, and this is undesirable if we need sign in guards to work
+    # with existing but unauthenticated user objects.
+    #
+    # DeferredSignInUser wraps a user object and can be used as such, and its
+    # present? method returns true only if the user is present AND password is
+    # correct.
+    #
+    # @return [Boolean]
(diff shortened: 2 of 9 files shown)


derekprior:
I agree that this should be easier to implement than it currently is. I haven't had the time to really dig into this and review it, but my first thought is that I wonder if we could handle this without a specific feature opt-in. Can we improve current guards and the sign in process to make this work.

I wonder if instead of sending `nil` into the stack when we have a valid email but invalid password if we should instead pass in a special `Guest` or `UnauthenticatedUser` object or something (haven't thought too much about this clearly) that would have the email address associated with it so you could look up the user if you wanted. Then we'd update sign in code to raise an error if it ever tried to sign in one of these objects.

I'll try to look at this on Friday if I get the time. It's possible by then I'll think the above idea is crazy.

thiagoa:
@derekprior I agree, changing the sign in process is generally a better idea. I was a bit conservative with this kind of change because I did not know if it was desired, so I did it as an opt-in. Also, it would be a bit of a breaking change.

This kind of change would have to go into a major version, right? I like your implementation idea, but with the following changes:

1. The `User#authenticate` method would always return a `SignInUser` object, **unless** the user can not be found in the database. `SignInUser` would be a `SimpleDelegator` to a `user` – it would store the incoming `password` param and would respond to a `signed_in?` method, which would verify the password validity. This would guarantee the `SignInGuards` to always receive a user object, if the email exists.
2. `DefaultSignInGuard` would call `current_user.signed_in?` to decide on the response: `SuccessStatus` or `FailureStatus`.
3. If a `SignInUser` gets past the sign in guards with an invalid password, the code would trigger an exception.

What do you think?

thiagoa:
@derekprior I forgot to say, given this change the `User#authenticate` method name would probably make no sense anymore. We could change this to something else.

derekprior:
I'm not inclined to merge this now as I think it complicates sign in a bit too much.

If I were going to implement this in my current application, I'd consider a couple of options. The first would be overriding `authenticate` in controllers to something like so:

```ruby
def authenticate(params)
  super || log_unsuccessful_attempt
end
```

`log_unsuccessful_attempt` would create an `unsuccessful_authentication` record in a database. Then I'd add a `SignInGuard` that would disallow sign in if there are a certain number of unsuccessful authentication records with a certain timeframe.

Alternatively, you could do something similar driven from the `SessionsController`:

```ruby
    sign_in(@user) do |status|
      if status.success?
        redirect_back_or url_after_create
      else
        log_unsuccessful_attempt(params)
        flash.now.notice = status.failure_message
        render template: "sessions/new", status: :unauthorized
      end
    end
```

PR #663, opened 2016-04-15
User proper base class for User model

derekprior:
In Rails 5, you are meant to have models that inherit from
`ApplicationRecord`. The code to inject clearance code into your
existing user model was hard-coded to look for a model that inherits
from `ActiveRecord::Base`. The specific inheritance requirement was
removed.

Similarly, applications that generated completely new user models were
set to inherit from `ActiveRecord::Base`. This should be
`ApplicationRecord` for Rails 5 and newer.

### DIFF
--- a/lib/generators/clearance/install/install_generator.rb
+++ b/lib/generators/clearance/install/install_generator.rb
@@ -24,10 +24,11 @@ module Clearance
           inject_into_file(
             "app/models/user.rb",
             "  include Clearance::User\n\n",
-            after: "class User < ActiveRecord::Base\n"
+            after: "class User < ",
           )
         else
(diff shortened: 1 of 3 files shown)


lib/generators/clearance/install/install_generator.rb
```diff
@@ -126,6 +127,14 @@ def migration_version
           "[#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}]"
         end
       end
+
+      def models_inherit_from
+        if Rails.version >= "5.0.0"
```
tute: I think both work well, but I'd probably write this as `ActiveRecord::VERSION::MAJOR >= 5`.

tute:
LGTM. 🚀

PR #666, opened 2016-04-22
Custom cookie name configuration spec

abunashir:
In [session_spec.rb:33](https://github.com/thoughtbot/clearance/blob/master/spec/clearance/session_spec.rb#L33)'s custom cookie name spec, we are setting up `cookie_domain ` configuration and that makes the test suite green. We have the `cookie_domain` configuraiton spec on [session_spec.rb:278] (https://github.com/thoughtbot/clearance/blob/master/spec/clearance/session_spec.rb#L278), so I belive this is to test `cookie_name` configuration.

This commit changes the spec to test out the cookie name has been set properly in the header

### DIFF
--- a/spec/clearance/session_spec.rb
+++ b/spec/clearance/session_spec.rb
@@ -32,12 +32,12 @@ describe Clearance::Session do
 
   context "with a custom cookie name" do
     it "sets a custom cookie name in the header" do
-      Clearance.configuration.cookie_domain = "custom_token"
+      Clearance.configuration.cookie_name = "custom_cookie_name"
 
       session.sign_in user
       session.add_cookie_to_headers(headers)
 
-      expect(headers["Set-Cookie"]).to match(/custom_token/)
+      expect(headers["Set-Cookie"]).to match(/custom_cookie_name=.+;/)
     end
 
     after { restore_default_config }


derekprior:
I was a bit confused by your description here. After looking at the change, you are basically making sure that the cookie name isn't something like "custom_token123" by adding the `=`. Is this correct?

abunashir:
Sorry to confuse you, Custom cookie name configuration spec passes for both of the following scenario:

- `Clearance.configuration.cookie_domain="custom_token"`, then header hash is

```ruby
{"Set-Cookie"=> "remember_token=7af2815d2161; domain=custom_token; path=/; expires=Sat, 22 Apr 2017 20:20:33 -0000" }
```

-  `Clearance.configuration.cookie_name="custom_token"` then header hash is

```ruby
{"Set-Cookie"=>"custom_token=e92ae87ba13400dff6067bca58c55b8fada24c11; path=/; expires=Sat, 22 Apr 2017 20:19:19 -0000"}
```

In first case we are setting up the custom_token as domain, but that's still passign the test, so I wanted to be more specific, that  `custom_token` is present as `custom_token=[token]` to make sure we are testing `cookie_name` configuration is working.

abunashir:
I just updated the commit to make the test less confusing :)

 `Clearance.configuration.cookie_name="custom_cookie_name"` instead of `custom_token`

spec/clearance/session_spec.rb
```diff
@@ -32,12 +32,12 @@
 
   context "with a custom cookie name" do
     it "sets a custom cookie name in the header" do
-      Clearance.configuration.cookie_domain = "custom_token"
```
abunashir: we are testing `cookie_name` configuration, but this was setting up `cookie_domain` configuration.

derekprior:
Ahh, I see now. Thanks

PR #672, opened 2016-04-29
Use flash alerts rather than notices

derekprior:
These messages are used to tell users they can't access a page without
signing in, that their password is incorrect, etc. These are much closer
to error or alert states than notice.


MattMSumner:
This seems like a good change. 👍

ledermann:
@derekprior Are there any plans to release 2.0? I would like to use the fixed flashes, but it seems they are stuck in the (stale?) 2.0 branch...

PR #677, opened 2016-05-04
Inject `include Clearance::User` inside model body

### LINKED ISSUE #676: Install Generator Broken by Commit 26f460d
[Commit 26f460d breaks the install generator](https://github.com/thoughtbot/clearance/blob/26f460db13e6657e43a4a60a583e77bd379e7775/lib/generators/clearance/install/install_generator.rb#L27). Broken in v1.14.0.

**Issue:** `include Clearance::User` is injected betweet `class User <` and `ActiveRecord::Base`. The generated code looks like this:

```ruby
class User <   include Clearance::User

ActiveRecord::Base

  # some other code
end
``` 

A fix should be to inject with a regex:

```
          inject_into_file(
            "app/models/user.rb",
            "\n  include Clearance::User\n\n",
            after: /class User < .*$/,
          )
 ```

The install generator spec probably should not only [check for the string](https://github.com/thoughtbot/clearance/blob/26f460db13e6657e43a4a60a583e77bd379e7775/spec/generators/clearance/install/install_generator_spec.rb#L40) to be present, but also check for proper file formatting.

abunashir:
Clearance generator was inserting `include Clearance::User` between `class User <` and `ActiveRecord::Base`, which is not correct.

Our spec was not checking if the generator was inserting clearance user model correctly or not. Let's add the test to ensure the generator is inserting the content properly. This commit fixes #676.

### DIFF
--- a/lib/generators/clearance/install/install_generator.rb
+++ b/lib/generators/clearance/install/install_generator.rb
@@ -24,7 +24,7 @@ module Clearance
           inject_into_file(
             "app/models/user.rb",
             "  include Clearance::User\n\n",
-            after: "class User < ",
+            after: "class User < #{models_inherit_from}\n",
           )
         else
           @inherit_from = models_inherit_from
--- a/spec/app_templates/app/models/rails5/user.rb
+++ b/spec/app_templates/app/models/rails5/user.rb
@@ -1,0 +1,5 @@
+class User < ApplicationRecord
+  def previously_existed?
+    true
+  end
+end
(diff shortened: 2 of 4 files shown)


abunashir:
The first build was passing on all version but Rails 5.0.0

Why:
Test helper `provide_existing_user_class`  using a template that inherits from `ActiveRecord::Base` all the time. In Rails 5.0 it should have created the User Class that inherits from  `ApplicationRecord` but it was not, and that was causing the suite to fail for version 5.0.

Solution:
I've added helper method `versionize_template` which can be used to get a version specific template when necessary. It will look for the template in the version-specific subdirectory.

Any suggestion?

derekprior:
Thanks for fixing this!

PR #678, opened 2016-05-05
Use config.redirect_url in deny_access matcher

christoomey:
The `deny_access` matched currently uses the hard coded path `/` for the
`singed_in?` case, but this does not match the implementation of the
`deny_access` helper, which uses `url_after_denied_access_when_signed_in` which
in turn uses the `Clearance.configuration.redirect_url`.

This change updates the `deny_access` matcher to also use the `redirect_url`.


derekprior:
Good catch. I'll cut a release with this shortly.

PR #680, opened 2016-05-10
Cleanup configuration spec

abunashir:
Remove unnecessary before hook and maintain consistency on setting up clearance configuration for each test example.

Why:

In the configuration spec, there are some before hook where we're basically setting up nothing, but defaults.

There is no consistency on setting up configuration, some example using before hook and others are not, some are using inline block and others are using multi-line block although it sets a single clearance configuration.

Let me know if it's okay then I can squash both of the commits into one.

### DIFF
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -4,48 +4,28 @@ describe Clearance::Configuration do
   after { restore_default_config }
 
   context 'when no user_model_name is specified' do
-    before do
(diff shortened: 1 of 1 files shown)


derekprior:
Thanks very much! I've been meaning to clean this up myself for a while.

abunashir:
Cool, I'm trying to manage some time to contribute every day, so If there is anything I can do to help, please let me know.

Issue #612, opened 2015-12-02
Should we store a digest of confirmation_token?


tute:
Do you think it's worth it to store a digest of the `confirmation_token`, so that access to the database doesn't grant easy access to user accounts? The problem is that it would be a backwards incompatible change, and would require a migration of existing data.

derekprior:
This is an interesting idea. I don't see why we would limit it to the confirmation_token. It seems the remember_token would be even more important to do this with.

One option would be to allow the values in the database to be stored as is, but expect that they are signed or encrypted with your app's secret key when passed from a user. This would be mostly backward compatible in that it would just invalidate existing sessions or outstanding password reset links.

This would prevent a leak of the database from exposing account access (not passwords - passwords are encrypted, of course). Presently your only recourse in this situation (as an app owner) is to rotate all remember tokens and confirmation tokens. If we made a change like this, you would need two pieces of information to make use of those tokens: the tokens themselves and the key with which they must be signed/encrypted in transit.

I'd be interested in other thoughts on this.

jferris:
If you use bcrypt, the encrypted values would be salted per-value, right? That would also let you leverage some of the existing code used for passwords.

derekprior:
Unless I'm missing something (quite possible), I don't think that would work. We could set their cookie to the the BCrypted value of the token that is stored in the database but anyone with access to the database could do the same thing. It's one more step someone with a database dump of accounts would have to go through, but it's just obscuring the token.

jferris:
We discussed this in person, and *I* was missing something. Derek is correct, and we need to handle this somewhat differently than we handle passwords.

monfresh:
I have come across a couple of projects that provide encryption and decryption of attributes:

https://github.com/attr-encrypted/attr_encrypted
http://rocketjob.github.io/symmetric-encryption/index.html

Thoughts on any of those?

PR #682, opened 2016-05-15
Add expiring, databaseless password reset tokens

### LINKED ISSUE #465: Expire password reset token after a period of time

derekprior:
It is a best security practice for password reset token to expire after
some amount of time. In Clearance 1.x, this was not the case. A password
reset email could be used months after it was originally sent so long as
no other password reset was ever completed.

In this change, password resets expire after 15 minutes (configurable)
or after the user successfully changes their password in any manner
(whichever comes first).

The token, confusingly called `confirmation_token`, is no longer stored
in the database. Instead, we use `ActiveSupport::MessageVerifier` to
sign the token and validate it when it is used. The message verifier is
configurable in case developers want to use something else.

In a future refactoring, I'd like to introduce a layer between Clearance
and `ActiveSupport::MessageVerifier` to make the API a bit more pleasant
to use, but this is an exercise for a future PR. For instance, I'd
prefer that the Clearance abstraction generate and validate tokens only
by taking a user object (and using the Clearance configuration).

Closes #465


derekprior:
Closes #465

app/mailers/clearance_mailer.rb
```diff
@@ -10,4 +12,14 @@ def change_password(user)
...
   end
+
+  private
+
+  def generate_password_reset_token(user)
+    Clearance.configuration.message_verifier.generate([
+      user.id,
+      user.encrypted_password,
+      Clearance.configuration.password_reset_time_limit.from_now,
+    ])
```
derekprior: I'm okay with this.

app/mailers/clearance_mailer.rb
```diff
@@ -10,4 +12,14 @@ def change_password(user)
       ),
     )
   end
+
+  private
+
+  def generate_password_reset_token(user)
+    Clearance.configuration.message_verifier.generate([
+      user.id,
```
derekprior: I'm also okay with this.

spec/controllers/passwords_controller_spec.rb
```diff
@@ -140,8 +156,16 @@ def update_parameters(user, options = {})
...
+      password_reset: { password: new_password },
     }
   end
+
+  def token_for(user)
+    Clearance.configuration.message_verifier.generate([
+      user.id,
+      user.encrypted_password,
+      15.minutes.from_now,
+    ])
```
derekprior: I'm okay with the way I have this broken up here.

spec/controllers/passwords_controller_spec.rb
```diff
@@ -140,8 +156,16 @@ def update_parameters(user, options = {})
...
-      token: user.confirmation_token,
-      password_reset: { password: new_password }
+      token: token_for(user),
+      password_reset: { password: new_password },
     }
   end
+
+  def token_for(user)
+    Clearance.configuration.message_verifier.generate([
+      user.id,
```
derekprior: I'm also okay with this.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -38,45 +37,50 @@ def update
...
     Clearance.configuration.user_model.
       find_by_normalized_email params[:password][:email]
   end
 
+  def find_user_by_password_reset_token(token)
+    verifier = Clearance.configuration.message_verifier
+    user_model = Clearance.configuration.user_model
+
+    begin
+      user_id, encrypted_password, expiration = verifier.verify(token)
```
derekprior: Maybe `encrypted_password` shouldn't be part of what we sign here? It allows us to invalidate any tokens as soon as the password is changed but it also means the encrypted password is part of the payload (signed). I dont know that the tradeoff is worth it. Curious for additional thoughts here...

jferris: What about storing a hash of the encrypted password? The token would still change with the password without leaking anything.

I can't think of reason you wouldn't provide a user with their encrypted password, but it makes me nervous anyway.

jferris:
This is a great idea.

README.md
```diff
@@ -56,6 +56,8 @@ Clearance.configure do |config|
   config.routes = true
   config.httponly = false
   config.mailer_sender = "reply@example.com"
+  config.message_verifier = ActiveSupport::MessageVerifier.new(secret_key_base)
+  config.password_reset_time_limit = 15.minutes
```
mike-burns: Nice.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -26,9 +25,9 @@ def edit
   def update
     @user = find_user_for_update
 
-    if @user.update_password password_reset_params
-      sign_in @user
-      redirect_to url_after_update
+    if @user.update_password(password_reset_params)
+      sign_in(@user)
+      redirect_to(url_after_update)
```
mike-burns: Yes!

app/controllers/clearance/passwords_controller.rb
```diff
@@ -38,45 +37,50 @@ def update
...
       find_by_normalized_email params[:password][:email]
   end
 
+  def find_user_by_password_reset_token(token)
+    verifier = Clearance.configuration.message_verifier
+    user_model = Clearance.configuration.user_model
+
+    begin
+      user_id, encrypted_password, expiration = verifier.verify(token)
+    rescue
```
mike-burns: What are you rescuing?

db/schema.rb
```diff
@@ -13,12 +13,11 @@
 
 ActiveRecord::Schema.define(version: 20110111224543) do
 
-  create_table "users", force: true do |t|
+  create_table "users", force: :cascade do |t|
```
mike-burns: What's this do?

spec/controllers/passwords_controller_spec.rb
teoljungberg: This seems like a bug... I wonder if Hound would be pleased by:

```ruby
Clearance.configuration.message_verifier.generate(
  [
    user.id,
    # ....
   ],
)
```

app/controllers/clearance/passwords_controller.rb
derekprior: The message verifier raises if the message is invalid. I should rescue that specific exception. That's part of why I want to insert a layer between Clearance and MessageVerifier. Rails made some improvements to the API in Rails 5 (it has a `valid?` method now).

At any rate, I will update to catch the right exception.

db/schema.rb
derekprior: It a change in how rails generates the schema in Rails 4+ that got picked up because this is the first time we've changed the test database schema in a long time!

geoffharcourt:
I had to implement expiring reset tokens in Clearance on a project last week. Very excited to see this become something upstream (and to remove dependency on the database).

tute:
This also fixes https://github.com/thoughtbot/clearance/issues/612. Thank you!

app/controllers/clearance/passwords_controller.rb
derekprior: I think it makes me nervous because I'd be providing anyone that can access that email access to their encrypted password. I'll play with it some to see what I can come up with.

Hashing the encrypted password is a decent idea. Complicates the lookup a tiny bit, but it may be worth it.

spec/clearance/password_reset_token_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    it "is nil if the signature is valid but expired" do
+      user = double("User", id: 1, encrypted_password: "foo")
+      token = Clearance::PasswordResetToken.generate_for(user)
+
+      Timecop.freeze(1.day.from_now) do
+        expect(token.user).to be nil
+      end
+    end
+
+    it "is nil if the signature is valid an unexpired by password changed" do
```
aripollak: s/an/and/, s/by/but/

lib/clearance/password_reset_token.rb
```diff
@@ -0,0 +1,43 @@
+module Clearance
+  class PasswordResetToken
+    def self.generate_for(user)
+      token = Clearance.configuration.message_verifier.generate([
+        user.id,
+        Digest::MD5.hexdigest(user.encrypted_password),
```
aripollak: This is a good idea, but doesn't it run the risk of exposing the existing encrypted password since MD5 isn't supposed to be particularly secure? That's a very edgy case, but it could theoretically be exploited in the same way that this PR is fixing? What about using the user's updated_at or a new field that keeps track of the last password reset time?

bdewater: Agree, avoid MD5 for any new code and use SHA256 instead. 

I'm wondering why this PR is only using signed messages and not encrypted and signed ones. An example of the latter is JSON Web Tokens (https://jwt.io/), those use HMAC-SHA256 and their usecase is nearly similar to this one.

robinvdvleuten:
Is there still any effort to get this into Clearance?

PR #683, opened 2016-05-25
Update README.md

JulienItard:
Add reference to clearance-i18n gem


derekprior:
I expanded on this a tiny bit and merged as abd6f75. Thanks

PR #686, opened 2016-07-05
remove unnecessary return from lib/clearance/user.rb

we138:
Hi! I was trying to use `clearance` and I found this small inaccuracy

**lib/clearance/user.rb**
```
def authenticate(email, password)
  if user = find_by_normalized_email(email)
    if password.present? && user.authenticated?(password)
      return user
    end
  end
end
```

I was decide corrected this, and make it in pr

### DIFF
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -135,7 +135,7 @@ module Clearance
       def authenticate(email, password)
         if user = find_by_normalized_email(email)
           if password.present? && user.authenticated?(password)
-            return user
+            user
           end
         end
       end


derekprior:
Thanks for the contribution.

I don't love the `return` there, but we also don't use inline conditionals in this code base as a matter of style. I'm also not bothered by the `return` enough to do something about it myself, but if you want to keep the current structure and change `return user` to `user`, I'd merge that, I suppose.

we138:
@derekprior ok, I remove inline conditional, but I'm curious why you are not using them?

ps
my question not for flame, I really wonder

derekprior:
It is part of the [thoughtbot Ruby style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby).

Personally, I prefer to see conditionals in block form as I feel it correctly identifies the complexity involved.  If you have two or three conditional blocks in a method, it looks bad but there's also a correspondingly high complexity to that method. You could make those three conditionals into one liners which would give the impression of a concise three-line method, but would actually hide the complexity in those conditionals.

we138:
@derekprior thanx, it is reasonable. By the way, I was remove inline conditional from my pr

we138:
hey @derekprior! I was changed pull request :muscle:

Issue #685, opened 2016-06-29
Incompatibility with the attr-encrypted gem


theharq:
Hi! I don't know if I should report this issue in the `attr-encrypted` project or maybe here is a good place.. in any case, there is an issue when we include the  `Clearance::User` module.

When we include this module the `PasswordStrategies::BCrypt` is included too by default, which adds a private method `encrypt` with one parameter:

 > ```
def encrypt(password)
   ::BCrypt::Password.create(password, cost: cost)
 end
```
https://github.com/thoughtbot/clearance/blob/master/lib/clearance/password_strategies/bcrypt.rb#L29

The `attr-encrypted` gem defines also a `encrypt` method but with 2 parameters which is called from this line:
> `send("#{encrypted_attribute_name}=", encrypt(attribute, value))`
https://github.com/attr-encrypted/attr_encrypted/blob/master/lib/attr_encrypted.rb#L159

Now if we try to update the object, we are going to see this error:

```
user.update(access_token: "test")
ArgumentError: wrong number of arguments (given 2, expected 1)
```

Thank you for your efforts and hard work maintaining this gem 😄

**Comment by tadp** [comment]
Steps to reproduce:

1. add `gem 'attr_encrypted'` to the Gemfile (notice the underscore)
2. add 
```   include Clearance::User
        extend AttrEncrypted
        attr_encrypted :token``` to your user.rb
3. `user.update(token: 'test')` (as above)

Notice this error:
```
ArgumentError: wrong number of arguments (given 2, expected 1)
from /Users/USER/.rvm/gems/ruby-2.3.1/gems/clearance-1.14.1/lib/clearance/password_strategies/bcrypt.rb:29:in `encrypt'
```

derekprior:
Closed by #693

PR #687, opened 2016-07-08
Fix incompatibility with the attr encrypted gem

we138:
Hi! I found this [issue](https://github.com/thoughtbot/clearance/issues/685), and I try to fix it in this way:
private method `encrypt` have a one line and is used once in the setter-method `password=()`. I suggest to remove method `encrypt` and used `::BCrypt::Password.create(password, cost: cost)` straight in to `password=()`.

### DIFF
--- a/lib/clearance/password_strategies/blowfish.rb
+++ b/lib/clearance/password_strategies/blowfish.rb
@@ -15,7 +15,7 @@ module Clearance
       #   gem
       def authenticated?(password)
         warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
-        encrypted_password == encrypt(password)
+        encrypted_password == _encrypt(password)
       end
 
       # @deprecated Use {BCrypt} or `clearance-deprecated_password_strategies`
@@ -26,14 +26,14 @@ module Clearance
         initialize_salt_if_necessary
 
         if new_password.present?
-          self.encrypted_password = encrypt(new_password)
+          self.encrypted_password = _encrypt(new_password)
         end
       end
 
       protected
 
       # @api private
-      def encrypt(string)
+      def _encrypt(string)
         generate_hash("--#{salt}--#{string}--")
       end
 
--- a/lib/clearance/password_strategies/sha1.rb
+++ b/lib/clearance/password_strategies/sha1.rb
@@ -16,7 +16,7 @@ module Clearance
       #   gem
       def authenticated?(password)
         warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
-        encrypted_password == encrypt(password)
+        encrypted_password == _encrypt(password)
       end
 
       # @deprecated Use {BCrypt} or `clearance-deprecated_password_strategies`
@@ -27,14 +27,14 @@ module Clearance
         initialize_salt_if_necessary
 
         if new_password.present?
-          self.encrypted_password = encrypt(new_password)
+          self.encrypted_password = _encrypt(new_password)
         end
       end
 
       private
 
       # @api private
-      def encrypt(string)
+      def _encrypt(string)
         generate_hash "--#{salt}--#{string}--"
       end
 
--- a/spec/password_strategies/bcrypt_spec.rb
+++ b/spec/password_strategies/bcrypt_spec.rb
@@ -69,6 +69,26 @@ describe Clearance::PasswordStrategies::BCrypt do
     end
   end
 
+  describe "#_encrypt" do
+    it "reserves #encrypt to prevent name conflicts with other gems" do
+      model_instance = fake_model_with_bcrypt_strategy
+
+      expect(model_instance.respond_to?(:encrypt, true)).to eq(false)
+    end
+
+    it "responds to #_encrypt" do
+      model_instance = fake_model_with_bcrypt_strategy
+
+      expect(model_instance.respond_to?(:_encrypt, true)).to eq(true)
+    end
+
+    it "ensures that #_encrypt is a private method as its name implies" do
+      model_instance = fake_model_with_bcrypt_strategy
+
+      expect(model_instance.respond_to?(:_encrypt)).to eq(false)
+    end
+  end
+
   def fake_model_with_bcrypt_strategy
     @model_instance ||= fake_model_with_password_strategy(
       Clearance::PasswordStrategies::BCrypt
--- a/lib/clearance/password_strategies/bcrypt.rb
+++ b/lib/clearance/password_strategies/bcrypt.rb
@@ -19,18 +19,17 @@ module Clearance
         @password = new_password
 
         if new_password.present?
-          self.encrypted_password = encrypt(new_password)
+          self.encrypted_password = _encrypt(new_password)
         end
       end
 
       private
 
       # @api private
-      def encrypt(password)
+      def _encrypt(password)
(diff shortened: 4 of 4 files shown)


magikid:
I don't think this is a good solution.  One of the reasons that the `encrypt` method is there is so that you can swap BCrypt out for a different strategy if you want.

we138:
@magikid Hi Chris, you absolutely right, I didn't noticed it. What do you think, how can fix it issue?

theharq:
Hello! and thank you for doing this first attempt to fix this method name collision. 💚 

Maybe I'm missing something but the method `encrypt` is not part of the public api of the module.

As the documentation says:

>   # You can supply your own password strategy by implementing a module that
>   # responds to the proper interface methods. Once this module is configured as
>   # your password strategy, Clearance will mix it into your Clearance User
>   # class. Thus, your module can access any methods or attributes on User.
>   #
>   # Password strategies need to respond to `authenticated?(password)` and
>   # `password=(new_password)`. For an example of how to implement these methods,
>   # see {Clearance::PasswordStrategies::BCrypt}.
> https://github.com/theharq/clearance/blob/master/lib/clearance/password_strategies.rb

So in our custom module as long we respond to `authenticated?(password)` and `password=(new_password)` and set this module in the configuration block.

```
Clearance.configure do |config|
  config.password_strategy = Clearance::PasswordStrategies::Custom
end
```

We could swap the default `PasswordStrategies::BCrypt`

This change may not be necessary for the other modules because all clearance-provided password strategies other than BCrypt have been deprecated.

Maybe we could update the [fake_model_with_password_strategy.rb](https://github.com/thoughtbot/clearance/blob/master/spec/support/fake_model_with_password_strategy.rb) to include the `encrypt` method so we make sure all the tests pass.. but I don't think the maintainers will want this.

Maybe there is another option to add a failing test for this edge case.

Thank you again, for taking the time to propose a solution 👍

magikid:
That's a good point @theharq.  I guess I should have read the comments a little better.

Because it's private api, I think a simple fix would be to just change the `encrypt` method name to something like `bcrypt_encrypt` (and changing it in the few spots that call `encrypt`).  Do you think that would be a decent solution?

theharq:
@magikid yeah that will be a good solution having a *strategy*`_encrypt` or `encrypt_password` method will do the trick, but still is up to the project maintainers and their plans for clearance 2.0

we138:
@theharq @mackuba hi guys, I have been renamed `encrypt` methods in whole strategies

mackuba:
Clearance? Hmm...



;)

tadp:
Thanks for tackling this.

I noticed this PR has no assignees but is in need of review. Maybe @derekprior can assign someone?

ACPK:
@magikid @tadp  - What is your recommended solution for this as of right now?

tadp:
@ACPK I like the renaming that @theharq suggested and that @we138 implemented in the latest commit.

`_` is [commonly used](https://www.quora.com/Why-using-underscore-to-represent-private-variable-method) in other languages to imply a private method and it is unlikely to be exposed by other gems

One thing I do see that this PR needs is a test to prevent this sort of thing from happening again by someone accidentally renaming the funny looking method back.

Thanks @we138 for doing this! WDYT about:
- [ ] adding this drop-in test to your PR that tests the method naming. It can be placed in `rspec spec/password_strategies/bcrypt_spec.rb:72`. I tried it out on your branch and it works:

```
  describe "#_encrypt" do
    it "reserves #encrypt to prevent name conflicts with other gems" do
      model_instance = fake_model_with_bcrypt_strategy

      expect(model_instance.respond_to?(:encrypt, true)).to eq(false)
    end

    it "responds to #_encrypt" do
      model_instance = fake_model_with_bcrypt_strategy

      expect(model_instance.respond_to?(:_encrypt, true)).to eq(true)
    end

    it "ensures that #_encrypt is a private method as its name implies" do
      model_instance = fake_model_with_bcrypt_strategy

      expect(model_instance.respond_to?(:_encrypt)).to eq(false)
    end
  end
```
- [ ] [Multiline commit message](http://chris.beams.io/posts/git-commit/) that explains the issue squashed into a single commit. This would make this really easy to revert or track down later.

I think with those two things this would be good for @derekprior to review.

magikid:
@acpk I believe the solution is to rename the `encrypt` method like this PR does.

we138:
@tadp yo Tad, your task is done!

tadp:
@we138 I think this is great. This PR is following the [contributing guidelines](https://github.com/thoughtbot/clearance/blob/master/CONTRIBUTING.md) well. However, I did not check off as completed "multiline commit message" in my previous comment because the commit message is still one line.

Maybe on the second line of the commit message it can say the WHY and HOW of the change?  It fixes incompatibility how?  It may be easy for us to see how it does now, but if someone does a git blame on the project months from now there is not much context on what is going on.

Here is the [thoughtbot styleguide](https://github.com/thoughtbot/guides/tree/master/style/git) on this. A short explaination should be fine.

derekprior:
Just wanted to note that I'm not ignoring this, but that I have had zero time for open source work of late. I will get to this, but it likely will not be until later this month. Thanks for understanding.

lib/clearance/password_strategies/blowfish.rb
```diff
@@ -15,7 +15,7 @@ module Blowfish
       #   gem
       def authenticated?(password)
         warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
-        encrypted_password == encrypt(password)
+        encrypted_password == _encrypt(password)
```
derekprior: The blowfish password strategy is deprecated. Please do not make any changes to this file. Thanks!

lib/clearance/password_strategies/sha1.rb
```diff
@@ -16,7 +16,7 @@ module SHA1
       #   gem
       def authenticated?(password)
         warn "#{Kernel.caller.first}: #{DEPRECATION_MESSAGE}"
-        encrypted_password == encrypt(password)
+        encrypted_password == _encrypt(password)
```
derekprior: The blowfish password strategy is deprecated. Please do not make any changes to this file. Thanks!

spec/password_strategies/bcrypt_spec.rb
```diff
@@ -69,6 +69,26 @@ def encrypted_password
     end
   end
 
+  describe "#_encrypt" do
```
derekprior: As long as the existing tests pass, I'm happy to skip specific unit tests for the private method.

derekprior:
This is a rather unfortunate side effect of password strategies being implemented as mixins to the User class rather than their own classes. 

Renaming `encrypt` solves this particular problem, but really we shouldn't be mixing in the methods at all. They are not useful to the user's `User` class. I opened a new PR where inline those methods: https://github.com/thoughtbot/clearance/pull/693

derekprior:
I went with the approach in #693 instead, but this is now fixed. Thanks so much to everyone involved in chasing this down.

I will try to release soon but I have spotty/unreliable wifi for the time being.

tadp:
Yeah sounds great.  Thanks @derekprior!! 💯 

And @we138 for the first pass, @theharq for bringing it up and the eyes of @magikid, @mackuba @ACPK.

derekprior:
1.14.2 is out with this fix. Thanks.

Issue #688, opened 2016-08-01
Why not use Clearance routes?


jjb:
the readme says:

> As of Clearance 1.5 it is recommended that you disable Clearance routes and take full control over routing and URL design.

As a brand-new Clearance user, I don't know the downfalls of using the default routes. What's the reason?

I'll make a PR for the readme with more detail.

Thanks!

magikid:
I believe the main reason behind that suggestion is so that if the gem's routes and URL design are changed, your website's URL design won't be affected when you update the gem.

jjb:
gotcha!

okay, made a PR

derekprior:
The idea was that it would insulate you from clearance changes and also prevents us from having to have configuration options for "I want this route to exist, but not this other one" or "I want the route to this action to be different."

Will clarify further in the doc PR.

PR #689, opened 2016-08-04
additional info about disabling Clearance default routes

jjb:
#688 

LMK if this isn't correct or complete and I'll update the PR.


magikid:
I think this looks good.

PR #690, opened 2016-08-08
Update README.md

vasinov:
Adding Hakiri badge to highlight some potential dependency security warnings (the way you have it in [Paperclip](https://github.com/thoughtbot/paperclip)).


derekprior:
Thanks for the contribution. I'm not inclined to merge this as I think the warnings it finds are misleading. For instance, the current build has 1 warning for code injection, which sounds really bad. Upon investigation, I see that it's for a nokogiri issue. I do not depend on nokogiri directly in this application. To solve this issue, I can either get everyone else (including Rails) to bump their acceptable version of nokogiri or I can create a false direct dependency on nokogiri.

I think these services are great for end user applications as the user is in the best place to decide how to mitigate any warnings.

vasinov:
@derekprior, yes, some dependencies are transitive and are not used in a gem directly. Those can be marked as false positives. Do you think there is still value in continuous security monitoring for non-transitive dependencies in this gem?

Issue #698, opened 2016-08-31
user forms on arbitrary pages


jjb:
There are some places in my app where I want to place a user registration or login form which will post to the regular built-in clearance controller.

The first problem I encountered when attempting this is a `nil` `@user` passed to the `form_for`. So I could copy-and-paste the [`@user` initialization code](https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/users_controller.rb#L13) into my other controllers but would rather not repeat the code.

Is there some sort of established way of achieving this?

When I get my answer I'll submit a PR with improved docs. Thanks!

pedrosmmoreira:
Hi @jjb, if I understand the issue correctly, you just want to render the the sign in/sign up form anywhere.

I would say the "established" way of doing this would be:

1) You can copy the views from clearance over to your application by running `rails generate clearance:views`. This will allow you to easily reference the form partials for rendering in your application.

2) You can wrap the form partials from Clearance with your own form builder, for instance:

For sign in:
```
<%= form_for :session, url: session_path do |form_builder| %>
  <%= render "sessions/form", form: form_builder %>
  <%= form_builder.submit %>
<% end %>
```

For sign up:
```
<%= form_for :user, url: users_path do |form_builder| %>
  <%= render "users/form", form: form_builder %>
  <%= form_builder.submit %>
<% end %>
```

Hope this helps!

jjb:
HI — thanks! Yes that's exactly what I'm talking about. I was mostly wondering about having to set `@user=User.new` (and anything else i was forgetting) in a controller. After I submitted the ticket I realized/remembered that I only need `:user` for the new form.

I'll put together a PR for the readme.

PR #700, opened 2016-09-05
notes on how to put forms into other views

jjb:
see discussion in #698


derekprior:
This seems like it's documenting either the contents of the views or how to render them? I'm not sure it needs mention in the README.

In fact, I think I'd rather update the `install` generator to copy views to your app (we currently have the `views` generator for this), which would make this more readily apparent.

Issue #701, opened 2016-09-06
rails d clearance:views deletes application.html.erb


lifeiscontent:
Yep.

pedrosmmoreira:
I think the issue to address here is if Clearance should even generate an application.html.erb file. I am thinking this was added as a convenience to help you get started, but since the suggested/necessary additions to layout are printed to STDOUT on install, it would be better to not touch the layout file.

I'll open a PR with a change in order to gather some opinions.

derekprior:
This is being discussed in #702. Closing this issue.

PR #702, opened 2016-09-06
Do not generate application layout file

pedrosmmoreira:
As mentioned in https://github.com/thoughtbot/clearance/issues/701, (which this solves), adding the application layout file to the list of views managed by Clearance can lead to unexpected/adverse effects.

While it is true that including the file with the suggested implementation of flashes and sign in/sign out logic is valuable in providing both a clear example of what is necessary to implement the library and in a good out-of-the-box experience, it can lead to inadvertently overriding the existing layout or, like in the issue mentioned above, have a surprising destructive effect when relying on
Rails generators to undo the views being copied.

Also, the very same implementation suggestion is presented to Stdout when running the installer, so this change does not imply the loss of any helpful input.

The PR moves the layout file to the testing application `spec/app_templates/testapp/app/views/layouts/application.html.erb` in order to maintain the acceptance tests passing as expected, and removes `application.html.erb` from the list of files managed under the Clearance generator.

### DIFF
--- a/spec/acceptance/clearance_installation_spec.rb
+++ b/spec/acceptance/clearance_installation_spec.rb
@@ -38,7 +38,6 @@ describe "Clearance Installation" do
     CMD
 
     FileUtils.rm_f("public/index.html")
-    FileUtils.rm_f("app/views/layouts/application.html.erb")
   end
 
   def testapp_templates
@@ -47,7 +46,6 @@ describe "Clearance Installation" do
 
   def configure_test_app
     FileUtils.rm_f("public/index.html")
-    FileUtils.rm_f("app/views/layouts/application.html.erb")
     FileUtils.cp_r(testapp_templates, "..")
   end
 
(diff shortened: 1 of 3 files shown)


magikid:
I like including some code to make it easy to get started. 

What about moving the header div into a partial and adding a little blurb at installation to include it in your application.html.erb?

derekprior:
Every time I've added Clearance to a project, I've already had a layout. I don't think it's valuable to include an application layout. The instructions after `rails generate clearance:install` even include an example snippet of how to set up your existing layout. Further, even if I didn't already have a layout, it's highly unlikely the generated layout would be what I wanted.

I'm inclined to merge this as is, but I'm curious if there are any other opinions here?

jferris:
Generating the header partial with the signed in/signed out logic might be nice. That seems tedious to write out for every application.

Having a line in the generator that checks to see if that partial is rendered from the layout might be nice, too, but is probably unnecessary.

pedrosmmoreira:
I am happy to add @jferris suggestions and repush. I agree with @derekprior, I really can't see a scenario where the layout isnt there. Thanks for the review and suggestions!

mjankowski:
Done via https://github.com/thoughtbot/clearance/pull/873

PR #709, opened 2016-10-06
make redirect to password redirect page dynamic, simply redirect to the current page

manuelmeurer:
After upgrading to v0.15 I ran into the problem that the URL helper for the "edit password" page in my app is not called `edit_user_password_url` but `edit_password_url`.
Since [the redirect](https://github.com/thoughtbot/clearance/blob/v1.15.0/app/controllers/clearance/passwords_controller.rb#L35) should always go the same page that is being requested right now, just without the param, I suggest to use `url_for` instead.

The tests pass. Not sure if another test is necessary?

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -32,7 +32,7 @@ class Clearance::PasswordsController < Clearance::BaseController
 
     if params[:token]
       session[:password_reset_token] = params[:token]
-      redirect_to edit_user_password_url(@user)
+      redirect_to url_for
     else
       render template: 'passwords/edit'
     end


app/controllers/clearance/passwords_controller.rb
```diff
@@ -32,7 +32,7 @@ def edit
 
     if params[:token]
       session[:password_reset_token] = params[:token]
-      redirect_to edit_user_password_url(@user)
+      redirect_to url_for
```
derekprior: I've not seen `url_for` used with no arguments before and I can't tell from first glance at the documentation what it's supposed to do. It it equivalent to `url_for :edit`, where `:edit` is the current action?

derekprior:
This was merged as de3233d and will be released in 1.15.1 shortly (there's an open PR for the release).

manuelmeurer:
Awesome, thanks!

coneybeare:
This merge is causing me problems in my app. I have my admin and public facing sides living int he same app, but all admin routes have a specific subdomain. When my user requests the change in password, my email forms correctly send them to the `edit_user_password_url`

    edit_user_password GET    /users/:user_id/password/edit(.:format)                                                                      admin/clearance/passwords#edit {:subdomain=>"MYSUBDOMAIN"}

but the `url_for` in this method does not look at subdomains at all, and redirects the user to the same path but without the subdomain, where they now encounter a routing error as it correctly should without the subdomain.

manuelmeurer:
Can you check if `redirect_to url_for(subdomain: request.subdomain)` works for you?

coneybeare:
It does, and this is how I've done it in my controller subclass' override, but is subdomain really the only edge-case here? When dealing with url's, shouldn't we stick to the pattern already established by this library? That is, configurable either by subclass-overridden method or within `Clearance.configuration`

manuelmeurer:
AFAICS subdomain is the only case that is not handled by `url_for`.

I think `url_for` is the best option here, compared to making this a config option, since it should work in all cases. Subclassing the controller is of course still possible if one wants completely different behaviour. 😄 

But the maintainers have the final say in this of course.

PR #714, opened 2016-10-24
Compliance with thoughtbot’s Ruby style guide

pablobm:
To keep our CI from hounding us. (Geddit?)


derekprior:
I think the existing version of the routes with the resource name on the first line and the options on following lines is more readable and more aesthetically pleasing than the updated version. I feel like it places the emphasis on the resource name which is appropriate. I reverted that change but will merge the others.

PR #720, opened 2016-12-15
Remove users/password#create route

danielnolan:
* There is a route for password#create nested in the users resource,
  and I can't figure out why you would need to create a password reset if
  you are logged in as a user. This seems like a useless route unless
  I am missing something.
* Remove the route for password#create under users
* Remove the route from the generator template

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,7 +13,7 @@ if Clearance.configuration.routes_enabled?
       only: Clearance.configuration.user_actions do
         resource :password,
           controller: 'clearance/passwords',
-          only: [:create, :edit, :update]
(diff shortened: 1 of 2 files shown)


danielnolan:
FWIW the build seems to be failing on ruby version 1.9.3. It looks like capybara has addressable as a dependency, but it allows any version `>= 0`. The most recent version of addressable requires ruby 2.0 or greater. Locking addressable to version `< 2.5` should fix this. I can open a separate pull request for that if you would like me to.

mjankowski:
Merged from https://github.com/thoughtbot/clearance/pull/870

PR #725, opened 2017-01-03
Reset Rails session on sign out

derekprior:
Signing out currently removes only the dedicated Clearance remember
token cookie. The Rails session itself is left intact because, as far
as Clearance is concerned, there is no user-specific information stored
there. This is true, but is also naive. It's common for Rails developers
to store user-specific information in `session`.

Resetting the session will also cause the CSRF token to be rotated. CSRF
issues typically involve signed-in users and Clearance sign out does not
leave the user signed in, so there's not much value to a CSRF attack
against a session in this state. Still, rotating the token is a best
practice.

### DIFF
--- a/lib/clearance/authentication.rb
+++ b/lib/clearance/authentication.rb
@@ -66,10 +66,14 @@ module Clearance
       clearance_session.sign_in user, &block
     end
 
-    # Destroy the current user's Clearance session.
-    # See {Session#sign_out} for specifics.
+    # Destroy the current user's Clearance and Rails sessions.
+    # See {Session#sign_out} for Clearance specifics. The Rails
+    # session itself is also reset as it is common practice for
+    # application developers to store user-specific information
+    # there.
     def sign_out
       clearance_session.sign_out
+      reset_session
     end
 
     # True if there is a currently-signed-in user. Exposed as a `helper_method`,
--- a/spec/controllers/sessions_controller_spec.rb
+++ b/spec/controllers/sessions_controller_spec.rb
@@ -82,6 +82,7 @@ describe Clearance::SessionsController do
         @user = create(:user)
         @user.update_attribute :remember_token, "old-token"
         @request.cookies["remember_token"] = "old-token"
+        session["test"] = "present"
         delete :destroy
       end
 
@@ -93,6 +94,10 @@ describe Clearance::SessionsController do
 
       it "should unset the current user" do
         expect(request.env[:clearance].current_user).to be_nil
+      end
+
+      it "should reset the rails session" do
+        expect(session["test"]).to be_nil
       end
     end
   end


sonalkr132:
I would suggest that session should be reset on login as well. Consider following test case:
```Ruby
setup do
  ActionController::Base.allow_forgery_protection = true
  @user = create(:user, handle: "bob", password: "lovesunicorns")
end

 test "authenticity token is reset on authentication" do
    get sign_in_path
    guest_csrf = request.session[:_csrf_token]

    # bob signs in
    post session_path(
      session: { who: "bob", password: "lovesunicorns" },
      authenticity_token: request.session[:_csrf_token]
    )

    # this will pass (consider a scenrio where bob uses a public computer where attacker
    # has got a valid csrf token from sign up/sign in page)
    patch "/profile", user: { handle: "bobistheworst"}, authenticity_token: guest_csrf
    assert_equal "bobistheworst", @user.handle

    # this will fail
    assert_raise ActionController::InvalidAuthenticityToken do
      patch "/profile", user: { handle: "alice" }, authenticity_token: guest_csrf
    end
  end
```

Here is devise [implementation](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/hooks/csrf_cleaner.rb#L1-L7) and [test](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/test/integration/authenticatable_test.rb#L330-L342), and sorcery [implementation](https://github.com/NoamB/sorcery/blob/80affcf2a892d9fe4213bca403b6fe197f67a234/lib/sorcery/controller.rb#L36) for protection against the same.

jferris:
When the user clicks "sign out," they're declaring "I no longer want to be recognized as this user," which is not necessarily the same thing as "I'm not using this computer anymore." If this gets merged, it will definitely introduce bugs into applications which assume the session doesn't get blown away in its entirety. Things like coupon codes and shopping carts often use the session, so this would be a breaking change.

I don't understand this scenario:

> bob uses a public computer where attacker has got a valid csrf token from sign up/sign in page

The risk vector is that somebody would get a CSRF token from a public computer, trick Bob into using that exact computer, and then trick Bob into submitting a form on another site which has that CSRF token embedded?

However, the CSRF token should be invisible to the user, so it seems safe to rotate that as often as we want. Is there any reason to wipe away the entire session and not just the CSRF token when the user logs in and out?

derekprior:
Upon some internal discussion we think that as written, this is likely to be a breaking change that may need to be put behind configuration. Resetting the entire session may not be a good idea.

Also, there's agreement upon our team that the idea that CSRF tokens should be "unique per user session" is troubling due to the overloaded term "user session". It can mean either:

1. The browser session (for instance, the length of the session cookie).
2. The period of time the user is signed in on the website.

My interpretation is that it is the former, rather than the latter, though I'm open to an explanation of how this is wrongheaded. Is there an actual vulnerability here as it relates to the CSRF token specifically?

Given:

1. A user signs in to a site using Clearance.
2. The user signs out of that site.
3. The users CSRF token is leaked (possibly via XSS?) to an external site

Then... ?

* What value is the CSRF token that isn't coupled with an authenticated user session?
* How is the value of this token increased by the fact that it *used to* be associated to an authenticated user?

derekprior:
> However, the CSRF token should be invisible to the user, so it seems safe to rotate that as often as we want. Is there any reason to wipe away the entire session and not just the CSRF token when the user logs in and out?

There's no public API for this that I've been able to find. From reading through devise, it seems like we can do something like what they do:

```ruby
warden.request.session.try(:delete, :_csrf_token)
```

I'm guessing that whatever middleware handles this, will see the missing token and regenerate it? We could do this as well (assuming thats how it works), but it does strike me as brittle.

sonalkr132:
Hi guys, I wish could provide satisfactory explanation to your queries like a security expert but sadly I can't. Hopefully @derekprior will forgive my repetitive arguments when I quote sources.

> The risk vector is that somebody would get a CSRF token from a public computer, trick Bob into using that exact computer, and then trick Bob into submitting a form on another site which has that CSRF token embedded?

Yes. I understand it is benign. I guess it is one half of session fixation attack.  [This blog](http://blog.plataformatec.com.br/2013/08/csrf-token-fixation-attacks-in-devise/) describes the scope as:
> The attack can only be exploited if the attacker can set the target session, either by subdomain cookies (similar to [described here](http://en.wikipedia.org/wiki/Session_fixation#Attack_scenarios)) or by fixation over the same Wi-Fi network. If the user knows the CSRF token, cross-site forgery requests can be made. More information can be [found here](http://homakov.blogspot.in/2013/06/cookie-forcing-protection-made-easy.html).

sonalkr132:
> What value is the CSRF token that isn't coupled with an authenticated user session?

I apologize that in my mail to Derek I started my description with logout when login/authentication step is more important. In my defense, thats how it was [report to us](https://hackerone.com/rubygems) and at that moment I had a very basic understanding of CSRF protection or session management. That's not to say I am an expert now :sweat_smile:

If your question had been *What value is the CSRF token in authenticated session that isn't coupled with a guest session?* I would say, if `csrf_token` is reset on authentication, attacker trying a session fixation will only have control over session id and hence won't be able to forge requests.

> 1. The browser session (for instance, the length of the session cookie).
> 2. The period of time the user is signed in on the website.
My interpretation is that it is the former, rather than the latter

[OWASP page](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Synchronizer_.28CSRF.29_Tokens) I used to quote that csrf token should be unqiue per user session describes csrf generation using [Synchronizer (CSRF) Tokens](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Synchronizer_.28CSRF.29_Tokens) as:

> In general, developers need only generate this token once for the current session. After initial generation of this token, the value is stored in the session and is utilized for each subsequent request until the session expires.

Unfortunately they didn't explicitly mention when the token is generated, however if you read further below whether they describe [Double Submit Cookie](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie) as alternative, it describes the same as:
> When a **user authenticates to a site**, the site should generate a (cryptographically strong) pseudorandom value and set it as a cookie on the user's machine separate from the session id.

I would say they mean authenticated user session while describing Synchronizer (CSRF) Tokens approach as well.

derekprior:
I still don't quite understand the session fixation entirely, but I am sufficiently convinced we should rotate the token on sign in and sign out. I do wish rails gave us public API for this, but I will probably go the route of Devise and just remove the value from the session.

I think this should still probably be behind a configuration for opt-in (at least in 1.0) as I wouldn't want this to break existing apps. I'm going to work on a new PR for this.

jferris:
@sonalkr132 I appreciate your patient and polite discussion on this issue. Security is always tricky, as it doesn't feel like many answers are obviously right, and most of us dealing with it are not really experts. It's much easier to work through issues like this when everybody takes a thoughtful approach. Thanks!

PR #728, opened 2017-01-06
Reset CSRF token on sign in

derekprior:
By rotating this token, we can prevent session fixation attacks. This is
the simplest explanation of session fixation that I've found, and I do
not want to recreate it here: https://www.owasp.org/index.php/Session_fixation

By rotating the CSRF token on sign in, the attacker will be unable to
make authenticated requests using the known session ID as they will not
have access to the new CSRF token without an additional security
vulnerability.

While I believe this should cause most Clearance users no harm if
enabled by default, there's a chance that it may, particularly if sign
in is done via AJAX, as additional form submissions after sign in will
not have the updated CSRF token.

For that reason, this feature is behind a configuration setting. The
configuration is `nil` by default, which will generate a warning that
encourages the user to set it to `true`, but the warning can be silenced
by setting the configuration to any value.

### DIFF
--- a/lib/clearance/authentication.rb
+++ b/lib/clearance/authentication.rb
@@ -62,8 +62,16 @@ module Clearance
     #
     # For an example of how clearance uses this internally, see
     # {SessionsController#create}.
+    #
+    # Signing in will also regenerate the CSRF token for the current session,
+    # provided {Configuration#rotate_csrf_token_on_sign_in} is set.
     def sign_in(user, &block)
-      clearance_session.sign_in user, &block
+      clearance_session.sign_in(user, &block)
+
+      if signed_in? && Clearance.configuration.rotate_csrf_on_sign_in?
+        session.delete(:_csrf_token)
+        form_authenticity_token
+      end
     end
 
     # Destroy the current user's Clearance session.
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -1,8 +1,6 @@
 require "spec_helper"
 
 describe Clearance::Configuration do
-  after { restore_default_config }
-
   context "when no user_model_name is specified" do
     it "defaults to User" do
       expect(Clearance.configuration.user_model).to eq ::User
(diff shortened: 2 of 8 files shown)


derekprior:
@jferris @sonalkr132 - attempt number two.

spec/configuration_spec.rb
```diff
@@ -146,4 +144,30 @@
...
+
+    it "is true and does not warn when `rotate_csrf_on_sign_in` is true" do
+      Clearance.configure { |config| config.rotate_csrf_on_sign_in = true }
+      allow(Clearance.configuration).to receive(:warn)
+
+      expect(Clearance.configuration.rotate_csrf_on_sign_in?).to be true
+      expect(Clearance.configuration).not_to have_received(:warn)
+    end
+
+    it "is false and does not warn when `rotate_csrf_on_sign_in` is  false" do
```
mwenger1: Extra space between words `is  false`

sonalkr132:
This looks good :+1: Thanks @derekprior!

lib/clearance/authentication.rb
```diff
@@ -62,8 +62,16 @@ def current_user=(user)
...
     # For an example of how clearance uses this internally, see
     # {SessionsController#create}.
+    #
+    # Signing in will also regenerate the CSRF token for the current session,
+    # provided {Configuration#rotate_csrf_token_on_sign_in} is set.
     def sign_in(user, &block)
-      clearance_session.sign_in user, &block
+      clearance_session.sign_in(user, &block)
+
+      if signed_in? && Clearance.configuration.rotate_csrf_on_sign_in?
```
jferris: Does it make more sense to do this in `clearance_session`? Currently, this method is a pretty straightforward delegator. Adding logic here means that you need both this module and the session object to actually sign in.

derekprior:
I think it's worth seeing what that looks like but I'm skeptical.

The two lines are related strictly to the rails session (distinct from the
clearance session). It would seem canonical for that code to belong in a
rails controller which is the scope this gets included in.

If we move it to Clearance session, then that code needs to gain the
ability to write to the rails session and to generate the new token, which
is currently only a method available on controllers (or in views via
helper_method).

PR #729, opened 2017-01-13
Remove unused nested `create` routes

abunashir:
In the routes file, we have two `create` routes defined for the `password` resource. One on the `password` resources and another one as nested `member` routes for users, but we only use the `resources` one to reset user's password.

This commit removes the unused nested `create` routes for the password resource.

Ref:

- https://github.com/thoughtbot/clearance/blob/master/config/routes.rb#L3-L5
- https://github.com/thoughtbot/clearance/blob/master/config/routes.rb#L16


derekprior:
duplicate of #720

PR #732, opened 2017-01-16
Avoid NoMethodError when receiving invalid params

### LINKED ISSUE #715: undefined method `[]' for nil:NilClass in Clearance::PasswordsController
Following passwords controller test fails with `NoMethodError: undefined method`[]' for nil:NilClass`:

``` Ruby
  context "when missing a parameter" do
    should "return bad request" do
      post :create
      assert_response :bad_request
    end
  end
```

It is coming from following [method accessing params](https://github.com/thoughtbot/clearance/blob/75e33bb0d0cd42aa97345d3496fd9e3e3ab73a05/app/controllers/clearance/passwords_controller.rb#L81) directly:

``` Ruby
  def find_user_for_create
    Clearance.configuration.user_model.
      find_by_normalized_email params[:password][:email]
  end
```

I have suggested following override in our app:

``` Ruby
  def find_user_for_create
    Clearance.configuration.user_model
      .find_by_normalized_email password_params[:email]
  end

  private

  def password_params
    params.require(:password).permit(:email)
  end
```

derekprior:
Previously, the `PasswordsController` assumed that well-formed input was
being passed to it via the password reset forms. While this is usually
the case, we can't control bad actors from probing however they'd like
and generating exceptions when this happens distracts from valuable
exception reports.

Fixes #715

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -31,7 +31,7 @@ class Clearance::PasswordsController < Clearance::BaseController
   def update
     @user = find_user_for_update
 
-    if @user.update_password password_reset_params
+    if @user.update_password(password_from_password_reset_params)
       sign_in @user
       redirect_to url_after_update
       session[:password_reset_token] = nil
@@ -48,8 +48,8 @@ class Clearance::PasswordsController < Clearance::BaseController
     mail.deliver_later
   end
 
-  def password_reset_params
-    params[:password_reset][:password]
+  def password_from_password_reset_params
+    params.dig(:password_reset, :password)
   end
 
   def find_user_by_id_and_confirmation_token
@@ -60,9 +60,13 @@ class Clearance::PasswordsController < Clearance::BaseController
       find_by_id_and_confirmation_token params[user_param], token.to_s
   end
 
+  def email_from_password_params
+    params.dig(:password, :email)
+  end
+
   def find_user_for_create
     Clearance.configuration.user_model.
-      find_by_normalized_email params[:password][:email]
+      find_by_normalized_email(email_from_password_params)
   end
 
   def find_user_for_edit
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,15 @@
 The noteworthy changes for each Clearance version are included here. For a
 complete changelog, see the git history for each version via the version links.
 
+## Unreleased
+
+### Fixed
+
+- Fix issue where invalid params could raise `NoMethodError` when updating and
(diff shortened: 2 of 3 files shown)


app/controllers/clearance/passwords_controller.rb
```diff
@@ -72,7 +72,8 @@ def password_reset_params
       ActiveSupport::Deprecation.warn %{Since locales functionality was added, accessing params[:user] is no longer supported.}
       params[:user][:password]
     else
-      params[:password_reset][:password]
+      reset_params = params[:password_reset] || {}
+      reset_params[:password]
```
MattMSumner: Ruby 2.3 introduces the [dig](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig) method that could be used here. I'm guessing clearance supports older versions of ruby though.

There is [a gem](https://github.com/Invoca/ruby_dig) that adds the method to older versions but I think this is fine as is.

app/controllers/clearance/passwords_controller.rb
```diff
@@ -85,8 +86,10 @@ def find_user_by_id_and_confirmation_token
   end
 
   def find_user_for_create
+    password_params = params[:password] || {}
```
MattMSumner: How would you feel about making this a method to use in both places? Maybe `password_params` or `email_from_params`?

app/controllers/clearance/passwords_controller.rb
derekprior: Yeah, we support back to 1.9.3, so that's not an option here. I don't want to take a dependency for it.

app/controllers/clearance/passwords_controller.rb
derekprior: The params are different in each place (`params[:password]` and `params[:password_reset]`). I thought about extracting the methods anyway (extracting two of them), as I don't usually like local variables that are easily avoided, but every method I add to a controller essentially becomes "public api".

app/controllers/clearance/passwords_controller.rb
```diff
@@ -45,7 +45,7 @@ def new
   def update
     @user = find_user_for_update
 
-    if @user.update_password password_reset_params
+    if @user.update_password(password_reset_params)
```
magikid: Since `password_reset_params` could return nil, would calling `@user.update_password(nil)` result in a user's password being set to `nil`?

mjankowski:
I rebased this, added some specs, and changed it to use `dig` (since we are ok re: ruby version support now).

Issue #726, opened 2017-01-04
why redirect with token in session?


jjb:
here in `PasswordsController`: https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/passwords_controller.rb#L33-L35

when the user comes in with their reset link from their email that has `token=a1b2c3d41` in it, the user is redirected back to the same page with that token in the session instead.

Why is this done? something to do with keeping tokens out of caches / histories?

It's a problem for me right now because my url for a user's password change is `/profile/password/edit?token=a1b2c3d4&user_id=123`, so this action redirects the user to a url that doesn't have the user_id. I'm going to have to monkeypatch that controller action to fix this. Not a big deal but wanted to check in on the motivation behind the behavior first.

Thanks!
John

----

vaguely related: https://github.com/thoughtbot/clearance/issues/615

jjb:
clearance by default requires user_id for looking up the user, even if/though the token is unique

https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/passwords_controller.rb#L79

there's nothing in the clearance user model or migration that makes confirmation token guaranteed to be unique, so this is in fact necessary.

yes, i'm easily overriding the controller action for my needs.

```ruby
  def edit
    @user = find_user_for_edit

    if params[:token]
      session[:password_reset_token] = params[:token]
      redirect_to url_for(user_id: params[:user_id])
    else
      render template: 'passwords/edit'
    end
  end
```

the addition is `(user_id: params[:user_id])`

what clearance proper can possibly do to accommodate my need and other future scenarios is strip token from GET params and add then into the redirect.

derekprior:
@jjb I can't think of a reason why there's no unique constraint on the indexes for `confirmation_token` and `remember_token`, but through normal use those tokens are generated via `SecureRandom` which is documented as suitable for "session key in HTTP cookies, etc". 

I'd still feel better if there were unique constraints though. PR welcome!

abunashir:
@derekprior I was going through the implementation for the password reset feature, and couple of things came in my mind. The first one is adding the `user_id` in the password reset link. I don't have a very good reason why not to use it but I was considering

- It's the only place we are exposing internal `user_id` that we used to store that specific user
- It might give a rough estimation about the `user base` for an application (how? Maybe create a new account then try to reset the password and there is the last id )

[...]

PR #734, opened 2017-02-08
Unique constraints for tokens

jjb:
inspired by discussion here: #726

The first commit to the migration simply adds unique constraints to the columns, which @derekprior  said he would invite (not sure if that meant for one or both token types)

The second commit takes advantage of the constraints, because user_id is now redundant.

I wasn't able to get my environment to run tests (ruby 2.4 problem?) so I just put this together for discussion, to see what changes are welcome.

If removing the user_id requirement isn't desirable, then the index in the first commit could be updated to be multi-column, `[:user_id, :confirmation_token]`.

LMKWYT!

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -76,12 +76,9 @@ class Clearance::PasswordsController < Clearance::BaseController
     end
   end
 
-  def find_user_by_id_and_confirmation_token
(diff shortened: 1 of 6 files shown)


dorianmariefr:
I think the idea was to add uniqueness for tokens without removing the need to be scoped to an user id (that has greater implications).

See #966

PR #740, opened 2017-03-14
Fixes install generator to create users table on connect exception

### LINKED ISSUE #737: Install fails if database not-yet-created
Running `bin/rails generate clearance:install` when a database has not yet been created results in the following error: 

```
rescue in connect': FATAL:  database "example_development" does not exist (ActiveRecord::NoDatabaseError)
```

Which traces back to this line in Clearance:
`lib/generators/clearance/install/install_generator.rb:105:in `users_table_exists?'`

This is sort of expected, but it would be nice if it were handled more elegantly. I can patch the `users_table_exists?` method to handle this case if someone can point me to an example of how Clearance prefers to write information back to the prompt.

ajm188:
This assumes that if we can't connect to the database, then it probably
hasn't been created. And, if it hasn't been created, then the users
table probably doesn't exist. So we should generate a migration to
create the table.

Resolves #737

### DIFF
--- a/lib/generators/clearance/install/install_generator.rb
+++ b/lib/generators/clearance/install/install_generator.rb
@@ -102,10 +102,16 @@ module Clearance
       end
 
       def users_table_exists?
-        if ActiveRecord::Base.connection.respond_to?(:data_source_exists?)
-          ActiveRecord::Base.connection.data_source_exists?(:users)
+        begin
+          connection = ActiveRecord::Base.connection
+        rescue ActiveRecord::NoDatabaseError
+          return false
+        end
+
+        if connection.respond_to?(:data_source_exists?)
+          connection.data_source_exists?(:users)
         else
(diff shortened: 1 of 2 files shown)


lib/generators/clearance/install/install_generator.rb
```diff
@@ -102,10 +102,16 @@ def migration_name_without_timestamp(file)
       end
 
       def users_table_exists?
-        if ActiveRecord::Base.connection.respond_to?(:data_source_exists?)
-          ActiveRecord::Base.connection.data_source_exists?(:users)
+        begin
+          connection = ActiveRecord::Base.connection
```
derekprior: I think extracting a `connection` method that returns this or raises a descriptive error on `NoDatabaseError` would be enough here.

In fact, even that may prove difficult because in rails 5 and newer the suggested command is `rails db:create` whereas in earlier version `rake db:create` should be run.

PR #743, opened 2017-05-03
Use standard paths for views and translations

tmsrjs:
I think this makes it easier to override the default views if you are already using `app/views/{sessions,users,passwords}/...` for some other controller and don't want/need to override the default Clearance controllers.
After this, you can simply put your views inside `app/views/clearance/{sessions,users,passwords}` and they will be picked up.

It also makes it more intuitive if your Clearance controllers are inside a module:
```ruby
# app/views/admin/sessions_controller.rb
class Admin::SessionsController < Clearance::SessionsController
  # will look for views in app/views/admin/sessions
  # or default to app/views/clearance/sessions
end
```
Translations will benefit from this too:
```erb
<%# app/views/admin/sessions.html.erb %>
<div id="clearance" class="sign-in">
  <%= t(".title") %>
  <%# is equivalent to %>
  <%= t("admin.sessions.title") %>
</div>
```
What do you guys think?


mbaird:
Ran into this same problem today, would be super useful to get this merged!

mjankowski:
Closing because this is quite out of date and I suspect has unrelated changes mixed in -- would be glad to review an updated version if still relevant!

PR #745, opened 2017-06-28
Remove unused method PasswordsController#url_after_create

alfius:
[This commit](https://github.com/thoughtbot/clearance/commit/5e107f6695461baa006cb03ed548255ecddd5a7c) changed the behavior to show a message instead of redirecting after changing the password, so the `url_after_create` method became unused.


mjankowski:
Done previously: https://github.com/thoughtbot/clearance/commit/ca48fd2db260d57598f508b88dc843a25353006e#diff-b25407c495af3eeed79d7f2615f8a764

Issue #719, opened 2016-12-14
Layout configuration via initializer deprecated?


joshuajansen:
It seems that the layout configuration via an initializer isn't supported anymore? (https://github.com/thoughtbot/clearance#layouts)

I can't get it to work with an out-of-the-box implementation of Clearance `1.15.1` And while looking through the latest source, I couldn't find any references to this configuration.

Is it deprecated or are there any prerequisites for this configuration that I could've missed?

Thanks!

derekprior:
This doesn't use any bit of Clearance, but rather uses a controller's public interface for changing the layout. You may need to restart your server after making this change?

joshuajansen:
Ah right, that makes sense. Not sure why this is not being picked up for my app then. Closing this, thanks!

michiels:
@joshuajansen Nice to see you here ;) Did you figure this out? Can't make it work myself.

joshuajansen:
@michiels I'm not 100% sure anymore, but I think it had something to do with the way I set up template inheritance. If I remember correctly, this was solved by not prepending the view path when the current controller is a Clearance controller.

michiels:
@joshuajansen ah ok thanks! Not doing template inheritance or view path modifying here. So maybe it's something else.

jvanbaarsen:
@joshuajansen / @michiels I think I found the solution. Instead of putting it in an initialiser you should put it in `config/application.rb`, then wrap it inside a `config.to_prepare do` block.

This is my `config/application.rb`:

```ruby
module MyApp
  class Application < Rails::Application
    config.load_defaults 5.1

    config.to_prepare do
      Clearance::PasswordsController.layout "login"
      Clearance::SessionsController.layout "login"
      Clearance::UsersController.layout "login"
    end
  end
end
```

I created a PR to update the README.

PR #746, opened 2017-07-04
Updated information about custom layouts

jvanbaarsen:
Setting the custom layout inside an initialiser does not work. Putting it in
your config/application.rb and then inside a `to_prepare` block does.

Fixes: #719


derekprior:
Thanks!

PR #748, opened 2017-07-19
	Don't declare helper_method if the method is not supported

Empact:
Eg. ActionController::API does not support helper_method


derekprior:
Thank you. Simplified and merged as 56443e6. I will release soon (hopefully by Friday at the latest).

PR #751, opened 2017-09-22
Correct typos in warning message

surreymagpie:
After installing Clearance on an app and adding a user, received message
about 'rotate_csrf_on_sign_in' being unset. I noticed two simple typos.
This commit simply corrects these and no functional change.


derekprior:
Thanks!

Issue #510, opened 2015-01-08
Use rails session instead of a cookie


arthurnn:
Why clearance use a separate cookie to save the token? any particular reason. 
Saving the current user id on rails session would probably make things easier. 
Thoughts?

derekprior:
What kind of things would be easier?

One reason I can think of not to do this is that currently clearances
cookie policy can be different than your rails session cookie.

jferris:
Session cookies are destroyed when the user closes their browser. The Clearance `REMEMBER_TOKEN` cookie needs to persist between sessions to remember the user.

jjb:
It's very non-standard to not use the built-in rails sessions, which also gives access to all the conventions and libraries that one might be accustomed to. It would be good to have some documentation about this in the readme — what the situation is, and what the rationale is. If that could be expressed here, I could put together a PR for the readme (and/or wiki documentation).

derekprior:
> It's very non-standard to not use the built-in rails sessions,

I'm sorry, but whose standard?

> It would be good to have some documentation about this in the readme — what the situation is, and what the rationale is.

I think Joe and I provided two decent examples here.

Are there specific problems you feel this creates?

jjb:
HI @derekprior, thanks for your quick response. Sorry if I sounded too complainy :)

> whose standard?

Yes, "standard" is perhaps an imprecise term. I meant to say:

* in my experience (using rails since 2007), it's uncommon for a library which is only used in rails and which needs to persist identity based on web browser instance to not use the built-in session management capabilities of rails.
* the benefits of using the built-in rails session management are:
 * confidence in the security model
 * familiarity with the configuration (e.g. configure rails session domain, https-only)
 * familiarity with the behavior (e.g. clearing rails session signs out the user)
 * ability to leverage other libraries (e.g. db-persisted sessions)


> currently clearances cookie policy can be different than your rails session cookie

I've never encountered a need for this but I suppose you and others have, so okay that's interesting. I see it as a drawback though, especially since clearance does use the rails session for some things. So one can't just turn off rails sessions entirely in order to simplify the configuration.

> Session cookies are destroyed when the user closes their browser. The Clearance `REMEMBER_TOKEN` cookie needs to persist between sessions to remember the user.

If the rails session `expire_after` configuration is given a value, then the cookie is not destroyed when the user closes their browser. I imagine @jferris is well aware of this though, so I'm not sure what he's referring to here.

jjb:
[...]

PR #753, opened 2017-10-05
Return authentication cookie only if authentication is performed

rogercampos:
It's not safe to unconditionally write the authentication cookie on
every response. If the application chooses to use http cache on a given
endpoint, and the first hit to that endpoint is performed by an
authenticated user, its cookie will be cached and reused as a caned
response for any other incoming request.

The authentication cookie should be set only if the application has used
authentication during this request.

### DIFF
--- a/lib/clearance/rack_session.rb
+++ b/lib/clearance/rack_session.rb
@@ -21,7 +21,11 @@ module Clearance
       session = Clearance::Session.new(env)
       env[:clearance] = session
       response = @app.call(env)
-      session.add_cookie_to_headers response[1]
+
+      if session.authentication_successful?
+        session.add_cookie_to_headers response[1]
+      end
+
       response
     end
   end
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -94,6 +94,13 @@ module Clearance
       ! signed_in?
     end
 
+    # True if a successful authentication has been performed
+    #
+    # @return [Boolean]
+    def authentication_successful?
+      !!@current_user
+    end
+
     private
 
     # @api private
--- a/spec/dummy/app/controllers/application_controller.rb
+++ b/spec/dummy/app/controllers/application_controller.rb
@@ -8,4 +8,12 @@ class ApplicationController < ActionController::Base
       render text: "", layout: "application"
     end
   end
+
+  def static_endpoint
+    if params[:use_current_user]
+      current_user # Just trigger the authentication
+    end
+
+    head :ok
+  end
 end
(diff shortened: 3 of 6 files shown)


spec/dummy/app/controllers/application_controller.rb
```diff
@@ -8,4 +8,22 @@ def show
...
+    if params[:use_current_user]
+      respond_raw_text("Current user is now #{current_user}")
+    else
+      respond_raw_text("static text")
+    end
+  end
+
+  private
+
+  def respond_raw_text(text)
```
derekprior: Can we respond with `head :ok` here and avoid rendering text at all? Doesn't look like the tests require it.

lib/clearance/rack_session.rb
```diff
@@ -21,7 +21,11 @@ def call(env)
       session = Clearance::Session.new(env)
       env[:clearance] = session
       response = @app.call(env)
-      session.add_cookie_to_headers response[1]
+
+      if session.send :authentication_performed?
```
derekprior: Is there a reason to use `send` here? Isn't this method public?

lib/clearance/session.rb
```diff
@@ -166,5 +166,9 @@ def cookie_value
 
       value
     end
+
+    def authentication_performed?
```
derekprior: Is this perhaps more accurately `authentication_successful`?

rogercampos:
Hi @derekprior thanks for taking a look at this. Sorry I don't know how other auth libs handle this, but if they use the rails session I guess they just avoid the problem altogether. If they don't manage cookies explicitly and instead rely on session, the application has also access to manage the session, and then it's possible for the app to avoid returning the session cookie on publicly cached endpoints (it's the app responsibility after all).

The problem with clearance is that it manages cookies explicitly and there's no way for the app to change this behavior per-endpoint.

(btw, I think clearance should manage their own cookie and not use session: https://github.com/thoughtbot/clearance/issues/510)

lib/clearance/rack_session.rb
rogercampos: I didn't know if you wanted to expand the public api of `Session`, if you're +1 to that I'll make it public instead

spec/dummy/app/controllers/application_controller.rb
rogercampos: yes, no need for that indeed, will change

mjankowski:
Merged via https://github.com/thoughtbot/clearance/pull/869

Issue #749, opened 2017-08-30
failing to reset your password breaks password reset indefinitely


eostrom:
Using Clearance 1.16.0 (and Rails 5.1.2, although I don't think it matters). Steps to reproduce:

1. Request a password reset.
2. Click the link in the email. You will see the new password form. DO NOT CHANGE YOUR PASSWORD.
3. Request a second password reset.
4. Click the link in the second email.

You should see the new password form again, but instead you will see the password reset email address form, with a message that says "Please double check the URL or try submitting the form again."

You can double check the URL and submit the form all you want, but password reset will remain broken until you clear cookies for the site.

(Step 2 – clicking the link but not bothering to fill out the form – is something of an edge case. We had a user who couldn't change her password due to an unrelated bug. But a browser could crash, a user could get distracted, etc. It shouldn't doom them forever.)

The cause of the bug: `PasswordsController#edit` stores the reset token in the session cookie. The token isn't deleted unless `PasswordsController#update` succeeds. And `find_user_by_id_and_confirmation_token` uses the token from the cookie, *even if there's a new token in the URL*.

A fix: Prefer the token in the URL over the one in the cookie. I believe this would fix all the use cases I've thought of, and I see no down side:

```
token = params[:token] || session[:password_reset_token]
```

gduquesnay:
This is happening on our app also.  Waiting on this issue to be fixed so we can update clearance.  I'm assuming 

```
token = params[:token] || session[:password_reset_token]
```

mentioned above is something to be changed inside clearance code.  Is there a workaround that we can use in the meantime while we wait for this to get patched by clearance?

abunashir:
@eostrom, @gduquesnay: I have created a PR with your suggestion and that should fix the issue, but in the meantime, you can also [override the password controller](https://github.com/thoughtbot/clearance#overriding-clearance) with the following code to fix the issue

```ruby
class PasswordsController < Clearance::PasswordsController
  private

  def find_user_by_id_and_confirmation_token
    user_param = Clearance.configuration.user_id_parameter
    token = params[:token] || session[:password_reset_token]
    
   Clearance.configuration.user_model.
      find_by_id_and_confirmation_token params[user_param], token.to_s
  end
end
```

PR #754, opened 2017-10-16
Fix password related issue for old/new token

abunashir:
Recently, to prevent password reset token via HTTP, there was an update to store the reset token in session and then redirect to the same edit page without token in the params, this works great but the way we find user introduced a new issue like 749

If the user requests a password reset and click on that link and for some reason that reset was unsuccessful and then the user request another link then it does not work and it keeps sending the user to the `password/new` page again and again.

Why is that happening?

The password `#edit`, and `#update` actions have before filter to ensure the token is valid for an existing user, and in the underneath `ensure_existing_user` is trying to find a user by the session token/parameter token.

But it gives precedence to the session token first and in above scenario, although the user has a new token in the URL, but the session still holds the old token which is not valid anymore and that creates the indefinite loop.

This commit adds a little tweak by switching the order and when a URL contains the `token` parameter then it updates the session reset token and then redirect to the same edit page. This way it updates the session & everything else stays same

Fixes: #749

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -78,7 +78,7 @@ class Clearance::PasswordsController < Clearance::BaseController
 
   def find_user_by_id_and_confirmation_token
     user_param = Clearance.configuration.user_id_parameter
-    token = session[:password_reset_token] || params[:token]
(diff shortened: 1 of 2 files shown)


derekprior:
@abunashir Thanks for the fix and the detailed commit!

PR #756, opened 2017-10-24
Extract factory module calculation to a private method

joshuaclayton:
What?
=====

With the rename of factory_girl to factory_bot, the sign_in helper
provided by Clearance needs to handle both possibilities.

This raises in the same manner as before when neither library is
present, but encourages factory_bot instead of factory_girl.


lib/clearance/testing/controller_helpers.rb
```diff
@@ -39,6 +37,21 @@ def sign_in_as(user)
...
         @request.env[:clearance].sign_out
       end
+
+      # Determines the appropriate factory library
+      #
+      # @api private
+      # @raise [RuntimeError] if both FactoryGirl and FactoryBot are not
+      #   defined.
+      def factory_module(provider:)
+        if defined?(FactoryGirl)
```
derekprior: Should these branches be swapped? As written, if I'm using FG 4.9 (when it's released), then I will get the deprecation warning, correct? I can avoid the deprecation in that case by using `FactoryBot`.

The user would still get the deprecation from the factory definition, which they own and control so will still know to switch to FB.

joshuaclayton: 👍 addressed in 5ed663ecc0bafd3e7e53acf7033d254f78e21177

PR #762, opened 2017-11-12
Add indexes to unique columns

IzumiSy:
This PR addes indexes for`email` and `remember_token` which are unique.


manuelmeurer:
I just found this PR while trying to figure out why there are no unique indexes on the `remember_token` and `confirmation_token` columns...
Does anyone know?

`email` is not necessarily unique, in our app the user is connected to a `userable` for example (admin, client etc.), so there might be duplicate emails (we're validating a unique email per userable class).

But the token columns should definitely be unique IMO, unless there is a specific reason against it...

PR #763, opened 2017-12-01
Do not depend on Rails

rolftimmermans:
Rails comes with Actioncable and all sorts of other stuff which not everyone might want to have installed. It's better to depend on the parts of Rails that are actually required by Clearance. Otherwise everyone is forced to install all parts of Rails instead of just the parts that are necessary.


Spone:
FYI this other PR seems to target the same goal: #776

mjankowski:
Thanks for this idea! Resolved by https://github.com/thoughtbot/clearance/pull/800

PR #769, opened 2018-01-19
Follow best practices for version comparison

pboling:
There are edge cases where string comparison of versions does not work.
As [this project/file is linked as an example](https://robots.thoughtbot.com/appraisal-find-out-what-your-gems-are-worth) of "the right way to setup Appraisals" it should exemplify those best practices.


derekprior:
Thanks!

PR #776, opened 2018-02-10
remove rails dependency

jjb:
This removes the "rails" gem as a dependency of clearance, to enable clearance users to require individual rails components instead of forcing them all as dependencies.

What motivated me to do this was the desire to not include actioncable in my project.

Let me know what you think!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,9 +2,12 @@ PATH
   remote: .
   specs:
     clearance (1.16.1)
+      actionmailer (>= 3.1)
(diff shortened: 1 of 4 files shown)


derekprior:
Thanks. This is a good idea!

derekprior:
One thing we need to do is generate the appraisal gemfiles again. I'll give that a try.

mjankowski:
Thanks for this idea! Resolved by https://github.com/thoughtbot/clearance/pull/800

jjb:
🆒

PR #780, opened 2018-03-09
Update Nokogiri development dependency

derekprior:
Nokogiri is just a development dependency, but we should still probably
not use a known-vulnerable version.


mjankowski:
I didn't see this before I made this - https://github.com/thoughtbot/clearance/pull/798 - but this one is a subset of that one, so alas, here we are.

PR #782, opened 2018-03-15
Allow user_parameter to be configurable.

lostmahbles:
I created a subclass of User with scoping that restricts the login/password reset ability of certain users. But doing this, changes the url parameter used in specs.

For example, with `config.user_model = ClearanceUser`, the routes expect users/:clearance_user_id/password.

In order to maintain the routing/naming structure of user_password_path, I needed to be able to force the named parameter to `:user_id`.

### DIFF
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -105,6 +105,11 @@ module Clearance
     # @return [Array<String>]
     attr_accessor :allowed_backdoor_environments
 
+    # The parameter for user routes. By default this is derived from the user
+    # model.
+    # @return [Symbol]
+    attr_accessor :user_parameter
+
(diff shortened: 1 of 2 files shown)


emptyflask:
Merge this PR, please. I'd like to be able to use a column other than "id" for my user routes.

mjankowski:
@lostmahbles this is pretty old, so sorry about late reply here.

Can you clarify a little what the scenario is that causes the problem this would fix?

There are a few places in controllers where we use either `user_parameter` or `user_id_parameter` to make sure we do a lookup on the right thing ... but in those cases it seems like having these reflect the configured overridden user model name would actually be desirable? I'm missing something here on why there's an issue.

PR #787, opened 2018-05-03
Allow BCrypt cost value to be configured

emptyflask:
The default value of 10 is no longer sufficient for some applications.
To calculate a cost factor based on how much time you'd like password
hashing to take (in milliseconds), see `BCrypt::Engine.calibrate(ms)`

To set your own cost value, use an inititalizer:
`BCrypt::Engine.cost = 12`

### DIFF
--- a/lib/clearance/password_strategies/bcrypt.rb
+++ b/lib/clearance/password_strategies/bcrypt.rb
@@ -6,6 +6,9 @@ module Clearance
     # will perform) is automatically set to the minimum allowed value when
     # Rails is operating in the test environment and the default cost in all
     # other envionments. This provides a speed boost in tests.
+    #
+    # To set your own cost value, use an inititalizer:
+    # `BCrypt::Engine.cost = 12`
     module BCrypt
       require 'bcrypt'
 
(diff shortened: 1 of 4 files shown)


emptyflask:
@derekprior I'm sure you're busy with other stuff, but would you mind taking a look at this PR? (assuming you're still the primary maintainer of Clearance)

lib/clearance/password_strategies/bcrypt.rb
```diff
@@ -19,18 +22,20 @@ def password=(new_password)
...
             cost: cost,
           )
         end
       end
+
+      def cost
+        if defined?(::Rails) && ::Rails.env.test?
+          ::BCrypt::Engine::MIN_COST
+        end
+        # Otherwise return nil, so that the cost stored in BCrypt
```
derekprior: What do you think about being explicit here and adding an else branch that return `BCrypt::Engine.cost`?

lib/clearance/password_strategies/bcrypt.rb
```diff
@@ -6,6 +6,9 @@ module PasswordStrategies
     # will perform) is automatically set to the minimum allowed value when
     # Rails is operating in the test environment and the default cost in all
     # other envionments. This provides a speed boost in tests.
+    #
+    # To set your own cost value, use an initializer:
```
derekprior: This assumes knowledge of BCrypt that most people won't have, so I'd think we should explain this value a bit more.

```suggestion
  # BCrypt has a `cost` argument which determines how computationally expensive
  # the hash is to calculate. The higher the cost, the harder it is for
  # attackers to crack passwords even if they posess a database dump of the
  # encrypted passwords. Clearance uses the `bcrypt-ruby` default cost except in 
  # the test environment, where it uses the minimum cost value for speed. If you
  # wish to increase the cost over the default, you can do so by setting a
  # higher cost in an initializer: 
```

mjankowski:
Merged in https://github.com/thoughtbot/clearance/pull/857

PR #788, opened 2018-05-17
Override default `:notice` flash key

### LINKED ISSUE #602: Flash :alert?
Just wondering why you guys are using `:notice` type flashes everywhere - even if it's a failure situation like here - `https://github.com/thoughtbot/clearance/blob/master/app/controllers/clearance/passwords_controller.rb#L86`?

In the app I'm working on, we're looking to give different colors for different types of flashes - green for success, etc.. So Clearance giving same key for all flashes looks weird :)

Would love to know why this is so and if there's a nice way to change this.

GrahamWalters:
Added a config option to override the default `:notice` key used for flash errors.

There's a fix in place on the 2.0 branch, but that branch hasn't seen any love in a while. This fix is backward compatible and allows the user to choose the key, which is useful if they're using Bootstrap and want to easily output `alert-danger`.

Fixes #602

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Clearance.configure do |config|
   config.secure_cookie = false
   config.sign_in_guards = []
   config.user_model = User
+  config.flash_error_key = :notice
 end
 ```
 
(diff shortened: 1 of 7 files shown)


mjankowski:
I'm going to close this b/c the change is in the `2.0.0.beta1` release and will be in 2.0 final at some point.

PR #799, opened 2018-10-22
Update ruby and rails support to remove versions past EOL

mjankowski:
I think this is the minimum set of changes needed to get ruby 2.3+ and rails 4.2+ included and specs passing without any deprecation output.

The dummy application config needed a few changes to silence deprecation warnings.

I flipped the behavior of the http shim - to be used only on rails less than 5, so that the majority of the Appraisal runs (5+) can use the default/newer style.


mjankowski:
I started this as a branch off of your "2.0" branch, thinking I'd continue that work ... but part way in realized that this might be fine against master.

It makes sense to me that 2.0 would be approximately this collection of versions in terms of support, but I'd also be fine calling this just an improvement in testing (at a minimum, it gets rails 5.1 and 5.2 into the matrix and will spot issues there...). I don't THINK there's anything here yet which would make a new 1.x release from master a problem.

derekprior:
> It makes sense to me that 2.0 would be approximately this collection of versions in terms of support, but I'd also be fine calling this just an improvement in testing (at a minimum, it gets rails 5.1 and 5.2 into the matrix and will spot issues there...). I don't THINK there's anything here yet which would make a new 1.x release from master a problem.

If you're going for 2.0 I'd probably just jump right to 5.2+, Ruby 2.4+ (maybe 2.5+) only at this point personally, but if you wanna keep 4.2 around that's probably fine too.

As is, I do think this would be fine a simplification for 1.x builds.

> Sort of related ... is it intentional that PRs don't generate travis builds?

Definitely not. Does it perhaps have to do with [this](https://status.github.com)? Those clowns at GitHub are still behind processing webhooks :trollface:. It's also possible something changed on the Travis side or with our config (unintentionally). It's really been a long time since I had done much with this repo.

mjankowski:
Sounds good ... I think I might push this in as a 1.x tidying, and then do a once over on open PRs and the 2.0 branch to see how things stand there.

PR #800, opened 2018-10-24
Remove rails as a dependency

mjankowski:
By depending directly on Rails, clearance becomes a source of dependency
inflation - pulling in every Rails framework into any app that uses clearance,
even when they were not needed.

This change updates the dependencies to just those which clearance actually
needs to run.

Inspired by priors - https://github.com/thoughtbot/clearance/pull/776 and https://github.com/thoughtbot/clearance/pull/763 - and updated against more recent changes.


jjb:
🆒

sidonath:
Hi @mjankowski! Are there any plans to build a new release with this change?

I see the work on 2.0 is underway, any chance of rolling out 1.6.2 before that? This shouldn't affect backwards compatibility, right?

mjankowski:
In theory the removal of the rails requirement should just be a silent change that doesn't impact anything ... I guess there could be someone who was unknowingly relying on clearance to pull in part of rails for them or something?

Either way - thanks for reminder. Yes, I'd like to do a release soon which gets out at least the stuff in master since last release, if not a full 2.0 rc release. Will review soon.

PR #802, opened 2018-10-25
Change rake db:migrate to rails db:migrate

jacobherrington:
Noticed this discrepancy following along with an Upcase course.


mjankowski:
Thanks.

PR #767, opened 2018-01-05
write a single cookie with the correct options

dholdren:
### Problem: httponly setting has no affect
I set `Clearance.configuration.httponly = true` but I noticed that executing `document.cookie` in Chrome's inspector still contained `remember_token`, then I realized that clearance was setting this cookie twice, once with `httponly`, and once without it.

Here's a spec that demonstrates the problem:
https://github.com/CasperSleep/clearance/commit/6f930bc235ed70f7b26750157ef66aaf7d89b30e

```
Failures:

  1) cookie options when httponly config value true sign in should have one remember_token cookie
     Failure/Error: expect(@remember_token_cookies.length).to eq(1), "expected one 'remember_token' cookie:\n#{@remember_token_cookies}"

       expected one 'remember_token' cookie:
       ["remember_token=1b3f09cead4c3ba39bed8ac0831837b0ea52a7b1; path=/; expires=Sat, 05 Jan 2019 18:04:16 -0000; HttpOnly", "remember_token=1b3f09cead4c3ba39bed8ac0831837b0ea52a7b1; path=/"]
     # /Users/dholdren/Dropbox (Personal)/third_party/clearance/spec/requests/cookie_options_spec.rb:41:in `block (4 levels) in <top (required)>'

  2) cookie options when httponly config value true sign in should have the httponly flag set
     Failure/Error: expect(@remember_token_cookies.last).to match(/httponly/i)

       expected "remember_token=fe6e5ed40284f7a724387fd25a2b2524918a9279; path=/" to match /httponly/i
       Diff:
       @@ -1,2 +1,2 @@
       -/httponly/i
       +"remember_token=fe6e5ed40284f7a724387fd25a2b2524918a9279; path=/"

     # /Users/dholdren/Dropbox (Personal)/third_party/clearance/spec/requests/cookie_options_spec.rb:45:in `block (4 levels) in <top (required)>'

  3) cookie options when httponly config value false sign in should have one remember_token cookie
     Failure/Error: expect(@remember_token_cookies.length).to eq(1), "expected one 'remember_token' cookie:\n#{@remember_token_cookies}"

       expected one 'remember_token' cookie:
       ["remember_token=5323f9bdb472031440d20a7283672d6590f856e8; path=/; expires=Sat, 05 Jan 2019 18:04:16 -0000", "remember_token=5323f9bdb472031440d20a7283672d6590f856e8; path=/"]
[...]

Issue #766, opened 2018-01-05
clearance writes two different cookies, clobbering settings like `httponly`


dholdren:
### Problem: httponly setting has no affect
I set `Clearance.configuration.httponly = true` but I noticed that executing `document.cookie` in Chrome's inspector still contained `remember_token`, then I realized that clearance was setting this cookie twice, once with `httponly`, and once without it.

Here's a spec that demonstrates the problem:
https://github.com/CasperSleep/clearance/commit/6f930bc235ed70f7b26750157ef66aaf7d89b30e

```
Failures:

  1) cookie options when httponly config value true sign in should have one remember_token cookie
     Failure/Error: expect(@remember_token_cookies.length).to eq(1), "expected one 'remember_token' cookie:\n#{@remember_token_cookies}"

       expected one 'remember_token' cookie:
       ["remember_token=1b3f09cead4c3ba39bed8ac0831837b0ea52a7b1; path=/; expires=Sat, 05 Jan 2019 18:04:16 -0000; HttpOnly", "remember_token=1b3f09cead4c3ba39bed8ac0831837b0ea52a7b1; path=/"]
     # /Users/dholdren/Dropbox (Personal)/third_party/clearance/spec/requests/cookie_options_spec.rb:41:in `block (4 levels) in <top (required)>'

  2) cookie options when httponly config value true sign in should have the httponly flag set
     Failure/Error: expect(@remember_token_cookies.last).to match(/httponly/i)

       expected "remember_token=fe6e5ed40284f7a724387fd25a2b2524918a9279; path=/" to match /httponly/i
       Diff:
       @@ -1,2 +1,2 @@
       -/httponly/i
       +"remember_token=fe6e5ed40284f7a724387fd25a2b2524918a9279; path=/"

     # /Users/dholdren/Dropbox (Personal)/third_party/clearance/spec/requests/cookie_options_spec.rb:45:in `block (4 levels) in <top (required)>'

  3) cookie options when httponly config value false sign in should have one remember_token cookie
     Failure/Error: expect(@remember_token_cookies.length).to eq(1), "expected one 'remember_token' cookie:\n#{@remember_token_cookies}"

       expected one 'remember_token' cookie:
       ["remember_token=5323f9bdb472031440d20a7283672d6590f856e8; path=/; expires=Sat, 05 Jan 2019 18:04:16 -0000", "remember_token=5323f9bdb472031440d20a7283672d6590f856e8; path=/"]
[...]

PR #804, opened 2018-10-25
Only set one cookie

mjankowski:
This is basically a reformulation of https://github.com/thoughtbot/clearance/pull/767 which exists to address https://github.com/thoughtbot/clearance/issues/766. It might be related to https://github.com/thoughtbot/clearance/issues/616 as well, but I haven't reviewed that thoroughly.


dholdren:
why not just comment on #767 with a suggestion to rebase and address whatever concerns you had?

PR #806, opened 2018-10-28
Add a check for the presence of generated code

jacobherrington:
It is possible to run the generator twice and append include
Clearance::User twice. This change keeps that from happening by checking
to see if Clearance::User has already been included in the file.

I'm not sure on the best way to write a test around this because it requires
the user to run the generator twice. I would love some feedback on testing
this change. Thanks!

### DIFF
--- a/lib/generators/clearance/install/install_generator.rb
+++ b/lib/generators/clearance/install/install_generator.rb
@@ -20,7 +20,10 @@ module Clearance
       end
 
       def create_or_inject_clearance_into_user_model
-        if File.exist? "app/models/user.rb"
+        if (File.exist? "app/models/user.rb") &&
+           (File.readlines("app/models/user.rb")
+                .grep(/include Clearance::User/)
+                .size.zero?)
           inject_into_file(
             "app/models/user.rb",
             "  include Clearance::User\n\n",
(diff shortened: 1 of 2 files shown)


lib/generators/clearance/install/install_generator.rb
```diff
@@ -20,7 +20,10 @@ def inject_clearance_into_application_controller
       end
 
       def create_or_inject_clearance_into_user_model
-        if File.exist? "app/models/user.rb"
+        if (File.exist? "app/models/user.rb") &&
+           (File.readlines("app/models/user.rb")
```
mjankowski: Maybe extract the content check part of this to a separate well named method - `user_model_includes_clearance?` (or something)?

mjankowski:
> I'm not sure on the best way to write a test around this because it requires the user to run the generator twice

There is an acceptance spec where we run a full installation, and part of that involves basically just running command line steps. You might want to look there for a start and see if running the generator twice from in there will get you the state you want to verify that this change works.

mjankowski:
I think that in the time since this was opened, it's now the case that the generator itself actually checks for dupe content and we don't need to have a custom solution for this.

If that's not the case can you provide a way to reproduce the issue? I ran `rails g clearance:install` a few times in a row, and it seemed to only insert the include line on the first run, and then not add dupes. This might be contingent on a certain rails version being used.

PR #808, opened 2018-10-29
Properly delete cookies with custom domain

rtymchyk:
This is primarily based on the conversation in https://github.com/thoughtbot/clearance/issues/616

Will be working on adding tests.

### DIFF
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -81,10 +81,15 @@ module Clearance
       end
 
       @current_user = nil
-      cookies.delete(
-        remember_token_cookie,
-        domain: Clearance.configuration.cookie_domain,
-      )
+
+      if Clearance.configuration.cookie_domain
+        cookies.delete(
+          remember_token_cookie,
+          domain: Clearance.configuration.cookie_domain,
+        )
+      else
+        cookies.delete remember_token_cookie
+      end
     end
 
     # True if {#current_user} is set.
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -81,7 +81,10 @@ module Clearance
       end
 
       @current_user = nil
-      cookies.delete remember_token_cookie, domain: Clearance.configuration.cookie_domain
+      cookies.delete(
+        remember_token_cookie,
+        domain: Clearance.configuration.cookie_domain,
+      )
     end
 
     # True if {#current_user} is set.
--- a/lib/clearance/session.rb
+++ b/lib/clearance/session.rb
@@ -56,15 +56,9 @@ module Clearance
       status = run_sign_in_stack
 
       if status.success?
-<<<<<<< HEAD
         # Sign in succeeded, and when {RackSession} is run and calls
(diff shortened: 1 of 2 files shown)


lib/clearance/session.rb
```diff
@@ -54,7 +54,9 @@ def sign_in(user, &block)
       status = run_sign_in_stack
 
       if status.success?
-        cookies[remember_token_cookie] = user && user.remember_token
+        cookies[remember_token_cookie] = cookie_options.merge(
+          value: user && user.remember_token,
+        )
```
rtymchyk: This merge ensures that we are maintaining the custom domain configuration that may have been passed when the user logs in. Without this merge, `sign_in` and `add_cookie_to_headers` end up creating 2 different cookies, one with the custom domain, and one without.

lib/clearance/session.rb
```diff
@@ -77,7 +79,14 @@ def sign_out
       end
 
       @current_user = nil
-      cookies.delete remember_token_cookie
+      if Clearance.configuration.cookie_domain.present?
+        cookies.delete(
+          remember_token_cookie,
+          domain: Clearance.configuration.cookie_domain,
+        )
```
rtymchyk: The API states that when you want to delete a cookie with a custom domain, you must pass that same domain option to it.

mjankowski:
It'd be good to see a test here - and it might be the case that the changes from https://github.com/thoughtbot/clearance/pull/804 have the side effect of fixing this cookie domain issue as well?

rtymchyk:
@mjankowski Yes, I believe your changes will fix the issue of duplicate cookies being written! But unfortunately it does not address the piece necessary to properly delete those cookies on sign out. I suppose the mixture of our changes would actually fix all issues here.

I'll be working on some specs during free time.

lib/clearance/session.rb
mjankowski: Is there an adverse effect to just always passing the domain here? If the value is nil ... what happens?

rtymchyk: That's a good point, I'll investigate.

mjankowski:
The other changes are merged here - if you want to rebase this to only cover the deletion scenario and add some specs (and confirm it's still an issue).

lib/clearance/session.rb
```diff
@@ -81,7 +81,15 @@ def sign_out
       end
 
       @current_user = nil
-      cookies.delete remember_token_cookie
+
+      if Clearance.configuration.cookie_domain
+        cookies.delete(
+          remember_token_cookie,
+          domain: Clearance.configuration.cookie_domain,
```
rtymchyk: Test case confirms that we cannot simply delete with `domain: nil` and must not specify the cookie option all together for cases without a custom domain.

rtymchyk:
@mjankowski I've rebased and added specs. The duplicate cookie creation is now fixed with your earlier PR, so this branch just tackles the deletion now.

rtymchyk:
Thanks for the merge @mjankowski

Issue #829, opened 2019-04-09
Allow Clearance::BackDoor in development environments


nsuchy:
I noticed the commit https://github.com/thoughtbot/clearance/commit/372c1514b892815914535d462409a354e1540964 restricts usage of `Clearance::BackDoor` to unit testing. Could this be modified to allow use in development environment, while this creates a security flaw in a production environment, it'd be useful for quickly switching between test accounts in development mode and is worth considering (as long as its kept off by default). Would a pull request be accepted for this change?

mjankowski:
Yes, discussed here as well - https://github.com/thoughtbot/clearance/issues/825

I could see either of these approaches probably working:

- Expand the list of envs where its allowed to be any of: `test`, `development`, `ci`
- Allow configuration of allowed envs, default to those

nsuchy:
I like the second approach better as it gives the developer more control. Could we just ask the developer to pass an array into the initializer and if `empty` or `nil` default to `test` and `development` and `ci`?

> Yes, discussed here as well - #825
> 
> I could see either of these approaches probably working:
> 
> * Expand the list of envs where its allowed to be any of: `test`, `development`, `ci`
> * Allow configuration of allowed envs, default to those

mjankowski:
Fixed via https://github.com/thoughtbot/clearance/pull/830

Issue #825, opened 2019-04-03
1.16.2 Broke tests in ci environment


A-Scott-Rowe:
The check in the BackDoor code for `test` environment broke tests in the `ci` environment. Could there be a way to specify test environments? Perhaps with an environment variable.

mjankowski:
Probably!

Just to be clear - you are running your tests on CI in an environment called `ci` - and thus the backdoor feature being disabled there (b/c it's not literally called `test`) is causing the problem?

I think your suggestion of expanding this check to be something like "if we are in the `test` environment OR `ENV['ENABLE_BACKDOOR']` is set to true, allow the backdoor" - might work. Is that what you meant?

A-Scott-Rowe:
Yes, that's exactly right. 

I've currently got a monkey patch in the initializer as a workaround. 

Also, after thinking about it for a day or two an even better idea may be to add a new attribute to `Clearance.configure`. That should make it more clear what environments are explicitly for testing.

mjankowski:
Fixed via https://github.com/thoughtbot/clearance/pull/830

PR #830, opened 2019-04-10
Fix Issue #829 (Allow Clearance::BackDoor in development environments)

nsuchy:
This pull request merges a fix which fixes Issue #829. The developer can now specify an array of allowed environments to allow Clearance::BackDoor. I modified existing test cases to account for the changes I made (perhaps I should add more tests?).

For further context see: https://github.com/thoughtbot/clearance/issues/829

### DIFF
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -31,8 +31,9 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      unless ENV["RAILS_ENV"] == "test"
-        raise "Can't use backdoor outside test environment"
+      @allowed_envs = Clearance.configuration.allowed_envs || ["test", "ci", "development"]
+      unless @allowed_envs.include? ENV["RAILS_ENV"]
+        raise "Can't use backdoor outside of explicitly allowed environments"
       end
 
       @app = app
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -31,9 +31,14 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      @allowed_envs = Clearance.configuration.allowed_envs || ["test", "ci", "development"]
+      @allowed_envs = Clearance.configuration.allowed_envs
+        || [
+              "test",
+              "ci",
+              "development"
+            ]
       unless @allowed_envs.include? ENV["RAILS_ENV"]
-        raise "Can't use backdoor outside of explicitly allowed environments"
+        raise "Can't use backdoor outside of explicitly allowed envs"
       end
 
       @app = app
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -31,12 +31,7 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      @allowed_envs = Clearance.configuration.allowed_envs
-        || [
-              "test",
-              "ci",
-              "development"
-            ]
+      @allowed_envs = Clearance.configuration.allowed_envs || ["test", "ci", "development"]
       unless @allowed_envs.include? ENV["RAILS_ENV"]
         raise "Can't use backdoor outside of explicitly allowed envs"
       end
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -31,7 +31,7 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      unless Clearance.configuration.allowed_backdoor_environments.include? ENV["RAILS_ENV"]
+      unless environment_is_allowed?
         raise "Can't use backdoor outside of explicitly allowed envs"
       end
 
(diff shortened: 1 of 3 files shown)


lib/clearance/back_door.rb
```diff
@@ -31,8 +31,9 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      unless ENV["RAILS_ENV"] == "test"
-        raise "Can't use backdoor outside test environment"
+      @allowed_envs = Clearance.configuration.allowed_envs || ["test", "ci", "development"]
```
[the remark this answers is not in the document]
mjankowski: We probably don’t need the default set here, it will already come from the configuration initialization.

mjankowski: If we set these defaults in the config itself, we probably don't need to set here?

lib/clearance/back_door.rb
```diff
@@ -31,8 +31,9 @@ module Clearance
...
+      unless @allowed_envs.include? ENV["RAILS_ENV"]
```
mjankowski: Can we skip the the i-var and use the config directly? Maybe wrap up the check in a well named private method?

lib/clearance/configuration.rb
```diff
@@ -90,6 +90,11 @@ class Configuration
     # @return [ActiveRecord::Base]
     attr_accessor :user_model
 
+    # The array of explictly allowed environments to use Clearance::BackDoor in.
+    # Defaults to ["test", "ci", "development"]
+    # @return [Array<String>]
+    attr_accessor :allowed_envs
```
mjankowski: Can we include "backdoor" in this config name so that it's clear what the envs are about?

nsuchy: See the new name set in my changes.

lib/clearance/back_door.rb
nsuchy: I agree. It now just checks the variable directly and assumes it's working as expected. Defaults were set in the configuration initializer anyways.

lib/clearance/back_door.rb
nsuchy: Agreed, resolved.

nsuchy:
Unless I'm missing something in the issue, this PR will also resolve https://github.com/thoughtbot/clearance/issues/825

lib/clearance/back_door.rb
```diff
@@ -31,8 +31,8 @@ module Clearance
   #   visit new_feedback_path(as: user)
   class BackDoor
     def initialize(app, &block)
-      unless ENV["RAILS_ENV"] == "test"
-        raise "Can't use backdoor outside test environment"
+      unless Clearance.configuration.allowed_backdoor_environments.include? ENV["RAILS_ENV"]
```
mjankowski: I realize it wasn't like this before you changed - but can we extract this to private method?

mjankowski:
I left a few what are probably final notes -- if you also want to a) rebase/squash this against master, b) add something to the NEWS file about this change - I'll merge in.

nsuchy:
For rebase/squash, I generally work on a branch and prefer to keep version history, unless there's a pressing reason to squash on my end can you use squash and merge?

mjankowski:
No prob - when your changes are good I can massage the commits to my liking.

lib/clearance/back_door.rb
nsuchy: Are you asking to make initialize a private method or the unless check a private method?

Additionally for the comment are you asking me to add extra unit tests to check different envs?

nsuchy:
@mjankowski Awesome, a commit history on a branch often makes it easier when going back to audit a pull request to see what the exact thought process was during development even if it does leave the PR itself a bit messy. I requested two clarifications and will make the changes after you reply 😄

lib/clearance/back_door.rb
mjankowski: Private method within `BackDoor` for the check - something like `unless allowed_backdoor_env?` or similar.

And yes - a unit test that shows the configuration overriding the default set of envs.

nsuchy: I see I'll make a private method to check the environment and call it instead.

I'm not too familiar with rspec though I came up with this based on the other tests? Would this work for our needs?
```ruby
it "can be used with non-default explicitly allowed envs" do
    Clearance.configuration.allowed_backdoor_environments.push("demo")
    with_environment("RAILS_ENV" => "demo") do
      user_id = "123"
      user = double("user")
      allow(User).to receive(:find).with(user_id).and_return(user)
      env = env_for_user_id(user_id)
      back_door = Clearance::BackDoor.new(mock_app)

      result = back_door.call(env)

      expect(env[:clearance]).to have_received(:sign_in).with(user)
      expect(result).to eq mock_app.call(env)
    end 
  end
```

nsuchy:
Okay so it appears that the unit tests were successful 😄

Is this ready for a merge or are there any final changes you'd like me to make?

mjankowski:
I rebased and did small cleanup, merged in here: https://github.com/thoughtbot/clearance/commit/b77d86622410a753ec3843333c29787845562c1b

While I was in there I:

- Added a case for totally disabling the backdoor by setting the config option to `nil`
- Added news file entry
- Changed error message to raise a message which includes the configured envs (easier debug if misconfigured).

Thanks!

PR #835, opened 2019-04-27
Do not test against Rails 4.2 if Ruby version is 2.6.x

carlesjove:
Rails 4.2 raises a ThreadError with Ruby 2.6 and higher, as documented in: https://github.com/rails/rails/issues/34790

This just a simple way to prevent tests from breaking if your current Ruby version is 2.6 or higher. I've noticed this combination is already allowed to break in Travis configuration file.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,9 +1,14 @@
 rails_versions = %w(
-  4.2
   5.0
   5.1
   5.2
 )
+
+if RUBY_VERSION < "2.6"
+  # Rails 4.x + Ruby 2.6 raises a ThreadError and tests will break
+  # https://github.com/rails/rails/issues/34790
+  rails_versions.prepend(4.2)
+end
 
 rails_versions.each do |version|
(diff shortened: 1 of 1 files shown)


mjankowski:
This makes sense.

The other thing I've thought about is just removing support for ruby 2.3 and rails 4.2 outright in the 2.0 release of clearance. Ruby 2.3 is already EOL and Rails 4.2 will not be maintained (I believe) after the release of 6.0 (soon).

carlesjove:
> The other thing I've thought about is just removing support for ruby 2.3 and rails 4.2 outright in the 2.0 release of clearance

In this case, may I suggest to add a (temporary) note in the README or CONTRIBUTING that Ruby < 2.6 has to be used in development, just to avoid other developers facing the `ThreadError` and having to search what the cause is?

mjankowski:
Wound up removing 4.2 - https://github.com/thoughtbot/clearance/pull/850

PR #844, opened 2019-06-27
Do not assume environment from env var

xymbol:
This prevents an exception from being raised when `RACK_ENV` is used to set the environment or, when `RAILS_ENV` is unset. Also, this makes Clearance work with Suspenders which uses `RACK_ENV` in newly generated apps.

https://github.com/thoughtbot/suspenders/blob/75158c86797b19324d80027677ec45cd88af2b02/templates/rails_helper.rb#L1


xymbol:
Wonder if someone is open to discuss this change. The current code is wrong and ignores the existing API. Proof: https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/railties/lib/rails.rb#L72-L74. /cc @mjankowski

mjankowski:
Thanks for ping.

Issue #842, opened 2019-06-05
Rails 6, custom user model and autoloading


matt17r:
I'm on Rails ~6.0.0rc2~ 6.0.0rc1 and Clearance 1.17.0 (and I'm using Minitest if that makes a difference).

I added `config.user_model = Person` to my Clearance initialiser but now when I run tests (using minitest) I get a deprecation warning that "Initialization autoloaded the constants ApplicationRecord and Person."

I also got that warning when I added `config.middleware.use Clearance::BackDoor` to my test environment configuration (as suggested by krsyoung in https://github.com/thoughtbot/clearance/issues/695), meaning I was getting the warning twice, but I've removed that for the time being.

Based on the warning in the docs (https://edgeguides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-and-initializers) I'm guessing they decided to start issuing a warning when initialisers trigger autoloading in Rails 6 but I'm not 100% sure how to check this.

It's also possible (and much more likely) that there is a bug in my code somewhere, rather than a conflict between this gem and Rails 6 but at this stage the app is very young and I don't think there's anything else that could be causing this.

mjankowski:
I don't actually see rails 6.0.0.rc2 released yet ... can you clarify that?

For what it's worth, I've been attempting to keep a [rails 6 prerelease branch](https://github.com/thoughtbot/clearance/tree/rails-six-prerelease) up to date, and will review there and attempt to replicate the reported error.

matt17r:
Whoops! You are of course right, 6.0.0rc1. Thanks for spotting that! :)

I found the error string in Rails: https://github.com/rails/rails/blob/master/railties/lib/rails/application/finisher.rb

Discussion about this change is in https://github.com/rails/rails/pull/35745

> Use cases include:
> [snip]
> - To set some class level configuration, e.g. SomeAPIServiceClass.backend = MyMockBackend. TBH I kinda hate that pattern as it trigger the eager loading of a ton of stuff, causing the app boot time to go though the roof. But at the same time I never figured any alternative pattern.

That's what Clearance is doing, if I understand correctly; letting the user specify a different class.

I was able to prevent the warning by requiring ApplicationRecord and my Person class at the top of the initialiser:

```
require 'application_record'
require 'person'

Clearance.configure do |config|
  # TODO Change the sender address
  config.mailer_sender = "reply@example.com"
  config.rotate_csrf_on_sign_in = true
  config.routes = false
  config.user_model = Person
end
[...]

PR #853, opened 2019-08-31
Support configuring custom user model with a string

odlp:
Resolves https://github.com/thoughtbot/clearance/issues/842

Since Rails 6.0 has switched to the Zeitwerk autoloader, referencing a model in an intializer triggers a deprecation warning:

```
Initialization autoloaded the constants ApplicationRecord and MyCustomUser
```

I've given this patch a try in a Rails 6.0 application which was displaying the deprecation warning and it resolved the issue.

### DIFF
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -112,7 +112,7 @@ module Clearance
     end
 
     def user_model
-      @user_model ||= ::User
+      (@user_model || "User").to_s.constantize
     end
 
     # Is the user sign up route enabled?
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Clearance.configure do |config|
   config.rotate_csrf_on_sign_in = false
   config.secure_cookie = false
   config.sign_in_guards = []
-  config.user_model = User
+  config.user_model = "User"
 end
 ```
 
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -8,9 +8,22 @@ describe Clearance::Configuration do
   end
 
   context "when a custom user_model_name is specified" do
-    it "is used instead of User" do
+    before(:each) do
       MyUser = Class.new
+    end
+
+    after(:each) do
+      Object.send(:remove_const, :MyUser)
+    end
+
+    it "is used instead of User" do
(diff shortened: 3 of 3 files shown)


lib/clearance/configuration.rb
```diff
@@ -112,7 +112,7 @@ def initialize
     end
 
     def user_model
-      @user_model ||= ::User
+      (@user_model || "User").to_s.constantize
```
odlp: I'm not sure of the performance impact of calling `constantize` repeatedly (because `Clearance` references `user_model` a lot).

Perhaps something a little less-concise would be better:

```diff
diff --git a/lib/clearance/configuration.rb b/lib/clearance/configuration.rb
index d7c6863..ee0d54e 100644
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -112,7 +112,11 @@ module Clearance
     end
 
     def user_model
-      (@user_model || "User").to_s.constantize
+      if @user_model.is_a?(String)
+        @user_model = (@user_model || "User").to_s.constantize
+      end
+
+      @user_model
     end
```

But in this version it feels awkward to be checking types.

odlp:
no-travis

I'm not sure why Travis decided not to run tests on this branch / PR...

lib/clearance/configuration.rb
rogercampos: clearance is already calling `constantize` on every request from `reload_user_model` in this same file. Since this is called from a `to_prepare` block it will be invoked every time in dev mode but only once in production, which is fine. 

I think `user_model` could be left as-is. The custom user class could be given as a string already, and `reload_user_model` will convert it into a class before each request.

The only necessary thing is make sure clearance doesn't use the `User` class during initialization.

odlp: That's an interesting suggestion - I tried it out quickly and it broke 48 / 198 examples - so I think the testing setup becomes a bit more involved (we'd have to call reload before each spec perhaps).

I do like the suggestion overall, but I think it's simpler to update `def user_model` because it's all self-contained - you can read it and understand what's going on, without having to know about the `reload_user_model` mechanism.

odlp:
@mjankowski I rebased on your updated `rails-six-prerelease` branch

PR #854, opened 2019-09-09
Support for the `SameSite` cookie attribute

artygus:
Same-site cookies is another layer to help prevent CSRF attacks. The feature is in [draft yet](https://tools.ietf.org/html/draft-west-first-party-cookies-06#section-4.1.1), but is already supported by modern browsers https://caniuse.com/#feat=same-site-cookie-attribute

### DIFF
--- a/spec/clearance/session_spec.rb
+++ b/spec/clearance/session_spec.rb
@@ -170,6 +170,31 @@ describe Clearance::Session do
     end
   end
 
+  context "if same_site is set" do
+    before do
+      Clearance.configuration.same_site = :lax
+      session.sign_in(user)
+    end
+
+    it "sets a same-site cookie" do
+      session.add_cookie_to_headers(headers)
+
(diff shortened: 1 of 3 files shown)


spec/clearance/session_spec.rb
```diff
@@ -170,6 +170,31 @@ def stub_status(success)
     end
   end
 
+  context "if same_site is set" do
+    before do
+      Clearance.configuration.same_site = :lax
```
mjankowski: Can this just be set to any non-nil value, or is there a collection of options?

artygus: there's a set of options `nil, :none, :lax, :strict` and a couple of aliases https://github.com/rack/rack/blob/6eda04886e3a57918ca2d6a482fda02a678fef0a/lib/rack/utils.rb#L232-L244,

sidonath:
Hi there :wave: Any chance of getting this PR merged? 😊

mjankowski:
Thanks for reminder.

PR #860, opened 2019-09-20
Release version 2.0.0

mjankowski:
I've updated a few small internal apps and the core functionality seems fine.

Some things we might want to document more thoroughly than current changelog:

- If you had specific styles for, or were in some way relying on the things clearance used the `notice` flash for, you'll have to change those to `alert` with 2.0.0
- The old password strategies have been deprecated - we might want to give people some links to other internet places where they can read about migrating through a password strategy change (ie, moving from SHA1 to BCrypt, etc)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    clearance (2.0.0.beta2)
+    clearance (2.0.0)
       actionmailer (>= 5.0)
       activemodel (>= 5.0)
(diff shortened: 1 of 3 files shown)


alex-benoit:
Hi @mjankowski - great work on this! Is there a timeline/ETA on this change?

mjankowski:
Thanks for ping -- I think it's ready to go but I've been strapped for time; I want to be sure we can support any post-release issues. That said ... it's been a while and we should probably just push it and do our best on that front.

Issue #839, opened 2019-05-06
EmailValidator does not single quotes in local part of email address in strict mode


HowardSnowden:
I noticed that the email validation fails for addresses like jerry'oconnell@somewhere.com.  From what I understand, single quotes are valid according to the spec.  

I was able to remedy my issue by forking and turning strict_mode to false in Clearance::User::Validations. I'm curious if:

a) Is this possibly  this is already a configuration option that I missed? 
b) If not, would it be a good idea allow this a configurable option? (I'm opening to doing the pull request myself).

PR #865, opened 2019-11-12
Update email_validator gem and relax validations

mjankowski:
The gem we use for email format validation recently changed its approach to
validation and radically relaxed the gem validation.

This blog post was part of the rationale: https://medium.com/hackernoon/the-100-correct-way-to-validate-email-addresses-7c4818f24643

This change updates the version of the gem we are using, and as a side effect,
relaxes the requirements for valid email addresses by default in clearance user
models.

Resolves https://github.com/thoughtbot/clearance/issues/839

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ PATH
       activemodel (>= 5.0)
       activerecord (>= 5.0)
       bcrypt (>= 3.1.1)
-      email_validator (~> 1.4)
+      email_validator (~> 2.0)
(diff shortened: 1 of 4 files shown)


mjankowski:
FWIW, I could be talked into either of "this relaxed approach sounds good, lets go with it" or "lets stand by what we were doing before".

If the former, this PR probably makes sense.

If the latter, we might want to switch to a gem who is taking a more strict approach, this one looks promising: https://github.com/afair/email_address

PR #784, opened 2018-04-10
Add parent_controller option to configuration

mgrachev:
Hi there!

We use `clearance` to authenticate admins in our application. For our purpose, we'll need to change the base class of `Clearance::BaseController` from `ApplicationController` to `Admin::ApplicationController`. Also, I saw the same issue #777.

So, I made the simple solution to this problem.

PR #867, opened 2019-11-13
Add parent_controller configuration option

mjankowski:
This uses https://github.com/thoughtbot/clearance/pull/784 as a starting point and then also:

- Does a rebase against changes in master since then;
- Makes small changes to avoid some deprecation warnings that would not be generated at the time the original PR was opened, around class names vs strings of class names


alex-benoit:
this is awesome!

PR #882, opened 2020-01-25
Deliver password reset email inline

croaky:
This commit changes the delivery behavior of the password reset email.
Before, it queued the email for later delivery.
Now, it delivers the email immediately.

The queue can be backed up, have no running workers, or other issues.
These issues can cause delay or no email delivery,
leading the user to abandon the site or contact support.
In the case of no delivery, there is also no error trace.

Inlining delivery also lets the programmer handle error cases such as
the email delivery service experiencing downtime or degraded service
or the email delivery service responding that it can't deliver the email
(such as a badly formatted or non-existent email address).
Handling these errors can help the user fix their own typos, etc.

Lastly, a project that integrates Clearance no longer needs to have
an ActiveJob-compatible background job process in their project.

https://api.rubyonrails.org/classes/ActionMailer/MessageDelivery.html#method-i-deliver_now

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,24 +285,6 @@ and `password` attributes. Over-riding the `email_optional?` or
 `skip_password_validation?` methods to return `true` will disable those
 validations from being added.
 
-### Deliver Email in Background Job
-
-Clearance has a password reset mailer. If you are using Rails 4.2 and Clearance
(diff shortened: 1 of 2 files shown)


croaky:
Sync'ed with upstream. Thoughts?

PR #817, opened 2019-01-22
Release version 2.0.0.beta1

mjankowski:
This is a rebase of the long-running but somewhat neglected `2.0` branch, which removes some previous deprecations, and updates the supported ruby/rails versions of clearance.

Opening this as a PR both to get feedback on changes in the branch, and also finalize the decisions around deprecation and version support.

PR #886, opened 2020-03-09
removed code trying to autoload files in deprecated path

udaykadaboina:
When Trying to use Clearance's [clearance-deprecated_password_stragegies](https://github.com/thoughtbot/clearance-deprecated_password_strategies)

If I change the password_stragegy to older methods, then the app would blowup. Since the Clearance Gem tries to autoload these files from within the gem, when these files were removed in this commit #817, specifically this [commit](https://github.com/thoughtbot/clearance/commit/be755401017f33f251f7c8ad76780e2d128a0f58#diff-ff65403a22f7c166322edeed5df872e2). 

`config.password_strategy = Clearance::PasswordStrategies::BCryptMigrationFromSHA1`


udaykadaboina:
@mjankowski Can you review this change please?

composerinteralia:
Thanks!

PR #889, opened 2020-03-24
Explicitly specify email uniqueness validator as case insensitive

unixmonkey:
Fixes deprecation warning from Rails 6.0.x versions:

> DEPRECATION WARNING: Uniqueness validator will no longer enforce case
sensitive comparison in Rails 6.1. To continue case sensitive comparison
on the :email attribute in User model, pass `case_sensitive: true`
option explicitly to the uniqueness validator.

Since the `case_sensitive` option has been supported on Rails since 1.x, this should be completely safe, while also being explicit enough for Rails 6+ to not throw any deprecation warnings.

### DIFF
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -143,7 +143,7 @@ module Clearance
         validates :email,
           email: { strict_mode: true },
           presence: true,
-          uniqueness: { allow_blank: true },
+          uniqueness: { allow_blank: true, case_sensitive: false },
           unless: :email_optional?
 
         validates :password, presence: true, unless: :skip_password_validation?


lib/clearance/user.rb
```diff
@@ -143,7 +143,7 @@ module Validations
         validates :email,
           email: { strict_mode: true },
           presence: true,
-          uniqueness: { allow_blank: true },
+          uniqueness: { allow_blank: true, case_sensitive: false },
```
sidonath, line 146: @unixmonkey @mjankowski I believe this was supposed to be `case_sensitive: true`.

This actually changes the logic quite a bit, since it builds the queries on `LOWER(email)` instead of `email`. This change broke our app because we didn't have an index on `lower(email)`, so our queries started timing out 😞 

This was originally changed in https://github.com/thoughtbot/clearance/commit/56f0f0489a138d64794136e133c5848183c576d5.

Also, the deprecation warning said we should set this to `true` if want to keep the behavior:

> To **continue** case sensitive comparison
> on the :email attribute in User model, pass `case_sensitive: true`
> option explicitly to the uniqueness validator.

This can make a big difference in performance and requires everyone to change their indexes, so it's not really a backwards compatible change. I'd suggest updating to `case_sensitive: true` and doing a bugfix release.

PR #893, opened 2020-04-25
Remove unused locale keys.

martinadamek:
It seems it's not used anywhere?
Looking for example at https://github.com/thoughtbot/clearance-i18n/blob/master/config/locales/de.yml confirms that, I guess?


martinadamek:
Test are failing, so closing ;-)

PR #894, opened 2020-05-02
Tiny changes to make clearance Neo4j-compatible

deobald:
I began this journey curious if Clearance would tolerate a Rails app backed by ActiveGraph instead of ActiveRecord. A quick grep of the Clearance codebase suggested it just might be possible... there is very little (if any) hard coupling to ActiveRecord.

These two tiny changes do not break ActiveRecord compatibility and don't alter the semantics of the queries they're performing.

`ActiveGraph` (previously `Neo4j.rb`) largely conforms to ActiveRecord interfaces but one of the things it lacks is the dynamically-defined `find_by_*` methods. However, it does contain `find_by(property:)` methods and changing these two lines makes Clearance work perfectly on a Neo4j.rb codebase.

I would love it if this change is acceptable to the thoughtbot folks so I can return to consuming the official gem. Please let me know if you have any questions. Thanks!

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -58,7 +58,7 @@ class Clearance::PasswordsController < Clearance::BaseController
     token = params[:token] || session[:password_reset_token]
 
     Clearance.configuration.user_model.
-      find_by_id_and_confirmation_token params[user_param], token.to_s
+      find_by(id: params[user_param], confirmation_token: token.to_s)
   end
 
   def email_from_password_params
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -121,7 +121,7 @@ module Clearance
       end
 
       def find_by_normalized_email(email)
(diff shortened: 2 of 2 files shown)


deobald:
@mjankowski Hi there. :) I just wanted to check in to see if this PR was at all within the realm of near-term possibility for a merge. I'd love to stop maintaining my branch and these two one-liner changes are so small they border on cosmetic.

I should note that I'm not intending to create any expectation of enforcing `ActiveGraph`-compatibility from the `clearance` end. It happens to be compatible because the ActiveGraph and ActiveRecord APIs are (intentionally) very close. Clearance is small and clean, which means that even if compatibility does break in the future, it's likely I can work around it from the outside.

Thanks so much!

mjankowski:
Sorry about delay, merged in now ... I'll push out release in near future after giving once over to other PRs and commits since last release.

deobald:
Not at all! Thanks so much for the merge. Our team really appreciates this!

PR #897, opened 2020-06-12
Add parameter stripping to BackDoor

purinkle:
Before, we were leaving the "as" parameter in the query string when accessing the back door. Doing so was causing problems downstream when the consuming application was using [Strong Parameters]. If the app had not listed the "as" parameter as a permitted parameter, then tests would fail. We added functionality to the back door to strip out the "as" parameter before calling the next middleware.

[Strong Parameters]: https://api.rubyonrails.org/classes/ActionController/StrongParameters.html

### DIFF
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -49,7 +49,8 @@ module Clearance
     # @api private
     def sign_in_through_the_back_door(env)
       params = Rack::Utils.parse_query(env["QUERY_STRING"])
-      user_param = params["as"]
+      user_param = params.delete("as")
+      env["QUERY_STRING"] = Rack::Utils.build_query(params)
 
       if user_param.present?
         user = find_user(user_param)
--- a/spec/clearance/back_door_spec.rb
+++ b/spec/clearance/back_door_spec.rb
@@ -50,12 +50,12 @@ describe Clearance::BackDoor do
     user_id = "123"
     user = double("user")
     allow(User).to receive(:find).with(user_id).and_return(user)
-    env = env_for_user_id(user_id)
+    env = build_env(as: user_id, foo: :bar)
     back_door = Clearance::BackDoor.new(mock_app)
 
     back_door.call(env)
 
-    expect(env["QUERY_STRING"]).to be_empty
+    expect(env["QUERY_STRING"]).to eq("foo=bar")
   end
 
   context "when the environments are disabled" do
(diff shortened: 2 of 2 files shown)


lib/clearance/back_door.rb
```diff
@@ -49,7 +49,8 @@ def call(env)
     # @api private
     def sign_in_through_the_back_door(env)
       params = Rack::Utils.parse_query(env["QUERY_STRING"])
-      user_param = params["as"]
+      user_param = params.delete("as")
```
mjankowski, line 52: If someone had an `as` param in a test, but was not using it for purposes of the clearance backdoor, would this have the unintentional side effect of the removing their param?

spec/clearance/back_door_spec.rb
```diff
@@ -46,6 +46,18 @@
...
+  it "strips 'as' from the params" do
+    user_id = "123"
+    user = double("user")
+    allow(User).to receive(:find).with(user_id).and_return(user)
+    env = env_for_user_id(user_id)
+    back_door = Clearance::BackDoor.new(mock_app)
+
+    back_door.call(env)
+
+    expect(env["QUERY_STRING"]).to be_empty
```
jferris, line 58: Will you also need to test for form data in POST requests? I don't think they'll end up in the query string.

lib/clearance/back_door.rb
georgebrock: Even before this change, passing an `as` parameter for other uses would have strange side effects. We'd be trying to use that param as a user ID before passing it through to the app, so it would either be inadvertently logging in as some random user if it happened to match a valid user ID, or it would be raising a record not found exception.

I guess if someone was passing a block that looks up the user in a way that doesn't raise (e.g. `config.middleware.use Clearance::BackDoor { |id| User.find_by(id: id) }`) then it wouldn't raise, but it would still unintentionally log them in.

Maybe it's worth allowing users to specify what this param is called to avoid collisions, but that's probably out-of-scope for this PR.

spec/clearance/back_door_spec.rb
georgebrock: The middleware always looks in the query string, even before this PR. Unless I'm misunderstanding what Rack puts in `env["QUERY_STRING"]` the middleware wouldn't find an `as` parameter in a POST body.

spec/clearance/back_door_spec.rb
```diff
@@ -46,6 +46,18 @@
     end
   end
 
+  it "strips 'as' from the params" do
```
georgebrock, line 49: Is it worth adding another test case here where we have multiple query params? I think the current test would pass if we deleted all the query params.

spec/clearance/back_door_spec.rb
jferris: From the [Rack SPEC], it looks like it won't put form-encoded data in `QUERY_STRING`, but looking at an example of using `as` in an application, I see that we're using it in the URL builder methods and not when building POST bodies, so this seems like a non-issue.

[Rack SPEC](https://www.rubydoc.info/github/rack/rack/file/SPEC)

spec/clearance/back_door_spec.rb
purinkle: Resolved in 52d7d94

georgebrock: Looks good to me!

dwightwatson:
Just noting that I believe this change breaks testing with nested attributes and RSpec - #913

PR #907, opened 2020-07-31
Separate out appraisal versions

composerinteralia:
Before this commit we were generation appraisals in a loop. This may
have worked fine when every version of Rails worked with the same set of
gems, but now that we need to conditionally constrain gem versions based
on the Rails version this no longer makes sense.

This commit declares each appraisal version separately, specifying only
the specific gem versions that are different from the top-level Gemfile
for that specific appraisal.

As an added bonus, moving some of these gems into the top-level Gemfile
made it possible to run the tests without appraisal as well.


Gemfile
```diff
@@ -5,10 +5,13 @@ gemspec
 gem 'addressable', '~> 2.6.0'
 gem 'ammeter'
 gem 'appraisal'
+gem 'capybara'
 gem 'database_cleaner', '~> 1.0'
 gem 'factory_bot_rails', '~> 5.0'
 gem 'nokogiri', '~> 1.10.0'
 gem 'pry', require: false
+gem 'rails-controller-testing'
+gem 'rspec-rails', '~> 4.0.0.beta3'
```
mjankowski, line 14: In future change, we should check if 4.0.0 is good.

composerinteralia: Good point. I'll follow up with a PR to bump some versions.

PR #909, opened 2020-08-02
Prevent user enumeration by timing attacks

### LINKED ISSUE #636: Make user enumeration by timing attacks more difficult
When authenticating a user, we skip any calls to BCrypt if we weren't able to find a user with the provided email address. This results in a faster response time to the client and less work on the server... but it also means one could possibly monitor the response times from the app to see if the provided user has an account.

When we detect that a user does not exist with the provided email, we should go ahead and perform a bcrypt check anyway, and always return false from that check. This will keep timings consistent with the response one gets when they use an email that exists but get the password wrong.

I took a quick look at how this could be done. I feel like the best way would be a new "dummy_authenticate" method on the password strategy that is called for this explict purpose. That's a breaking change to the contract of the password strategy though.

Another alternative would be to handle this in the user mixin itself. We'd want to make sure we produce similar timings in any event.

composerinteralia:
Closes #636

Before this commit, we were skipping any BCrypt (or other password
strategy) checks when we couldn't find the user by email. This opened up
the possibility for a bad actor to detect whether an account existed for
a given email address based on the timing of the response.

This commit mitigates the problem by creating a throw away user and
setting a dummy password on it, triggering BCrypt to create an encrypted
password.

The added tests were consistently failing before this change because
authentication for accounts that did NOT exists was taking 1-2ms longer
that for accounts that did exist (with the cost set to
::BCrypt::Engine::MIN_COST in tests). The timings are now very close.
The `be_within` delta could almost have been 0.0001, but I stuck with
1ms to avoid any flakiness in the test suite.

### DIFF
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -117,6 +117,8 @@ module Clearance
           if password.present? && user.authenticated?(password)
             user
           end
+        else
+          prevent_timing_attack
         end
       end
 
@@ -129,6 +131,13 @@ module Clearance
       end
 
       private
+
+      DUMMY_PASSWORD = "*"
+
+      def prevent_timing_attack
+        new(password: DUMMY_PASSWORD)
+        nil
+      end
 
       def password_strategy
         Clearance.configuration.password_strategy || PasswordStrategies::BCrypt
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -47,6 +47,35 @@ describe User do
       expect(User.authenticate(user.email, "bad_password")).to be_nil
     end
 
+    it "takes the same amount of time to authenticate regardless of whether user exists" do
+      user = create(:user)
+      password = user.password
+
+      user_exists_time = Benchmark.realtime do
+        User.authenticate(user.email, password)
+      end
+
+      user_does_not_exist_time = Benchmark.realtime do
+        User.authenticate("bad_email@example.com", password)
+      end
+
+      expect(user_does_not_exist_time). to be_within(0.001).of(user_exists_time)
+    end
+
+    it "takes the same amount of time to fail authentication regardless of whether user exists" do
+      user = create(:user)
+
+      user_exists_time = Benchmark.realtime do
+        User.authenticate(user.email, "bad_password")
+      end
+
+      user_does_not_exist_time = Benchmark.realtime do
+        User.authenticate("bad_email@example.com", "bad_password")
+      end
+
+      expect(user_does_not_exist_time). to be_within(0.001).of(user_exists_time)
+    end
+
     it "is retrieved via a case-insensitive search" do
       user = create(:user)
 


spec/models/user_spec.rb
```diff
@@ -47,6 +47,35 @@
       expect(User.authenticate(user.email, "bad_password")).to be_nil
     end
 
+    it "takes the same amount of time to authenticate regardless of whether user exists" do
```
gnfisher, line 50: Is running just one comparison enough to feel confident here? Or should we run it multiple times and make sure all of them remain within the expected delta?

lib/clearance/user.rb
```diff
@@ -130,6 +132,13 @@ def normalize_email(email)
 
       private
 
+      DUMMY_PASSWORD = "*"
+
+      def prevent_timing_attack
+        new(password: DUMMY_PASSWORD)
```
eebs, line 138: My understanding is that this will call `password=`, which will result in the password strategy running. For BCrypt that is

https://github.com/thoughtbot/clearance/blob/1dd09b0e51c839109ab4c342d0d38eeb39d27663/lib/clearance/password_strategies/bcrypt.rb#L22-L31

In the case where a user was found, we'd end up calling `authenticated?`, which would again call the password strategy

https://github.com/thoughtbot/clearance/blob/1dd09b0e51c839109ab4c342d0d38eeb39d27663/lib/clearance/password_strategies/bcrypt.rb#L16-L20

We want these two code paths to take the same amount of time to avoid leaking via timing attacks. Do `BCrypt::Password#==` and `BCrypt::Password.create` use the same process under the hood that causes them to take the same amount of time? Will this be true for all password strategies?

composerinteralia: For BCrypt the timings seem to be the same

```rb
require 'benchmark'
require 'benchmark/ips'
require "bcrypt"

encrypted_password = BCrypt::Password.create("password")

Benchmark.ips do |bm|
  bm.report("compare") { BCrypt::Password.new(encrypted_password) == "password" }
  bm.report("create") { BCrypt::Password.create("password") }
  bm.compare!
end
```

```
Warming up --------------------------------------
             compare     1.000  i/100ms
              create     1.000  i/100ms
Calculating -------------------------------------
             compare      4.889  (± 0.0%) i/s -     25.000  in   5.114814s
              create      4.870  (± 0.0%) i/s -     25.000  in   5.134016s

Comparison:
             compare:        4.9 i/s
              create:        4.9 i/s - 1.00x  (± 0.00) slower
```

composerinteralia: For argon2 as well:

```rb
require 'benchmark'
require 'benchmark/ips'
require "argon2"

encrypted_password = Argon2::Password.new.create("password")

Benchmark.ips do |bm|
  bm.report("compare") { Argon2::Password.verify_password("password", encrypted_password) }
  bm.report("create") { Argon2::Password.new.create("password") }
  bm.compare!
end
```

```Warming up --------------------------------------
             compare     1.000  i/100ms
              create     1.000  i/100ms
Calculating -------------------------------------
             compare      8.335  (± 0.0%) i/s -     42.000  in   5.044675s
              create      8.314  (± 0.0%) i/s -     42.000  in   5.057424s

Comparison:
             compare:        8.3 i/s
              create:        8.3 i/s - 1.00x  (± 0.00) slower
```

eebs: Awesome, thanks for the benchmark! I think this is the best we can do until something changes. 👍

composerinteralia: I went with this approach because it was the only way I could think of fix this for any password strategy. Another way to fix would be to build a DUMMY_ENCRYPTED_PASSWORD, but the password strategy doesn't currently have an API for doing that so that would only work for our officially supported password strategies.

eebs: Yeah, I think what you have here is excellent. If we find that some supported strategy in the future has different timings for creating and equality checks we can revisit this.

spec/models/user_spec.rb
composerinteralia: Good question. How many times do you think would be reasonable?

gnfisher: I honestly don't know if it is even necessary. I was just thinking of the fuzz  testing I've seen done in Elm apps.

composerinteralia: If I run each 100 times I can bump the `be_within` delta from 0.001 to 0.0001 (* 100). That is nice, but it also adds 0.4 seconds to our test suite. I am torn.

composerinteralia: I think I am going to leave it alone for now to keep the test suite as fast as possible. These tests in their current form are enough to catch a regression to the old behavior. We can always revisit this if we find that the 1ms delta is not sufficient.

composerinteralia:
@eebs @gnfisher I am thinking of sending out a patch release today with just this change (we can release the other bug fixes when we meet again next week).

eebs:
Sounds good. I'd love to pair on that with you if we can find some overlapping time.

gnfisher:
I would also join if the time works ou!

PR #921, opened 2020-12-04
Fix password reset URLs when using a custom model

### LINKED ISSUE #918: Changing user_model doesn't change routes in templates
We changed our application's Clearance model to `Foobar` in the initializer. This changes a lot of things, including `Clearance.user_id_parameter`. To properly pull this value down, we needed to update our routes. However, when we updated our routes, the URL helpers in the views (for resetting passwords, for example) still point to routes that work for `user` namespaces. It seems like these should be generated dynamically with the value returned by `Clearance.user_parameter`

eebs:
When using a custom `user_model` other than `User`, generating the
password reset link and attempting to reset the password would raise an
error. This is because the URL helpers were hardcoded to use the `User`
model, such as `edit_user_password_url`.

This commit uses Rails' [PolymorphicRoutes] to build the URL from the
passed in model. This will use the record's model to generate the URL,
allowing the URL to respect any custom model.

Fixes #918

[PolymorphicRoutes]: https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ complete changelog, see the git history for each version via the version links.
 ### Fixed
 
 - Support for accessing Rails 6.x primary_key_type in generator.
+- Fix password reset URLs when using a custom model
 
 ## [2.3.0] - August 14, 2020
 
--- a/app/views/clearance_mailer/change_password.html.erb
+++ b/app/views/clearance_mailer/change_password.html.erb
@@ -2,7 +2,7 @@
 
 <p>
   <%= link_to t(".link_text", default: "Change my password"),
-    edit_user_password_url(@user, token: @user.confirmation_token) %>
+    url_for([@user, :password, action: :edit, token: @user.confirmation_token]) %>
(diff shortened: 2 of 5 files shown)


mike-burns (commented):
Great catch.

I'm not entirely sure how to test this either, but it looks like we have a test that tries changing the `user_model`: https://github.com/thoughtbot/clearance/blob/master/spec/configuration_spec.rb#L22

Maybe we could add a test that does that and then tries to navigate or generate the HTML?

eebs:
Thanks for the suggestion @mike-burns. I was able to create a test for the mailer in eeb55ca. I'm trying to use the same approach with a controller test to ensure that the password reset form is rendered correctly when the user visits the app using the password reset link, but it's proving to be a little more difficult.

mike-burns:
That mailer test is excellent. Unfortunately I don't actually know enough about Clearance to have any hints for the controller test.

PR #933, opened 2021-02-26
Update gems versions

MottiniMauro:
Bumping versions of nokogiri and action view due to some security
vulnerabilities reported by dependabot.


gnfisher:
We decided to use dependabot to bump the gem versions.

Issue #929, opened 2021-01-30
Incompatible with Turbo


dwightwatson:
It appears that Clearance is incompatible with Turbo out of the box, as Turbo will no longer render responses to forms **unless** they have a 4xx/5xx status.

* `UsersController#create` returns `render template: "users/new"` on a validation error - this may need to be updated to `render template: "users/show", status: :unprocessable_entity`
* `PasswordsController#create` returns `render template: "passwords/create"` regardless of whether a password reset has been sent or not - this may need to be updated to a redirect somewhere.
* `PasswordsController#edit` returns `render template: "passwords/edit"`without a token on the request - this may need to be updated with `:unprocessable_entity` also.

Adding `status: :unprocessable_entity` is an easy fix for two of these, and I wonder if perhaps `PasswordsController#create` and just redirect back to `PasswordsController#new` with a successful flash message.

Related: #925

manuelmeurer:
Would love to see Turbo support added! I did the necessary changes in my fork: https://github.com/manuelmeurer/clearance/tree/turbo
Would a PR be acceptable? #936 is already open with most of the fixes.

PR #936, opened 2021-03-13
Update status codes on failures

dwightwatson:
This fixes one of the two issues that prevent Clearance from being compatible with Turbo. 

Simply - it updates the status for failed forms from 200 to 422. For context, Turbo will only render a 4xx/5xx response to a form. With the 200 response Turbo will do nothing, and the user will be left sitting on the form page.

I expect this would be a breaking change so should likely be slated for the next major release, along with the second fix for Turbo which is a little more involved.

Related: #929

### DIFF
--- a/spec/controllers/passwords_controller_spec.rb
+++ b/spec/controllers/passwords_controller_spec.rb
@@ -111,7 +111,7 @@ describe Clearance::PasswordsController do
           user_id: user,
         }
 
-        expect(response).to have_http_status(:unprocessable_entity)
+        expect(response).to be_successful
         expect(response).to render_template(:edit)
         expect(assigns(:user)).to eq user
       end
@@ -211,6 +211,7 @@ describe Clearance::PasswordsController do
         )
 
         expect(flash.now[:alert]).to match(/password can't be blank/i)
+        expect(response).to have_http_status(:unprocessable_entity)
         expect(response).to render_template(:edit)
         expect(cookies[:remember_token]).to be_nil
       end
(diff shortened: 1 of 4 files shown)


spec/controllers/passwords_controller_spec.rb
```diff
@@ -111,7 +111,7 @@
           user_id: user,
         }
 
-        expect(response).to be_successful
+        expect(response).to have_http_status(:unprocessable_entity)
```
MatheusRich, line 114: @dwightwatson Can you add a similar test for the users_controller as well?

dwightwatson: Looks like the test case was missing, so I've added it now.

MottiniMauro:
@dwightwatson Thanks for working on this! It looks like the CI caught a test failure in `spec/controllers/passwords_controller_spec.rb:106`, could you look into it?

dwightwatson:
Alright, think I've sorted the specs out. If you can trigger the CI again it should be good 🤞

manuelmeurer:
Can I help with anything to get this merged?
It seems like there are many more `render template:` calls that would need to be rendered with status 422?

dorianmariefr:
Looked good to me, I expect it to fix Turbo applications and break some tests that depend on the returned status code (but not break the application)

dorianmariefr:
@manuelmeurer I did the changes with tests in #965

I also tried to have a full turbo application generated and running tests but I think that was overkill

PR #944, opened 2021-07-19
Use `#dig` to get session data

sidonath:
If the sign in form was submitted without any params (e.g. by a bot or malformed request), the `#authenticate` method would raise the 'NoMethodError: undefined method `[]' for nil:NilClass' exception.

We can work around that with `ActionController::Params#dig` which would treat missing or misnamed params just like invalid credentials.


sidonath:
The test failure looks flakey

PR #945, opened 2021-08-31
Fix open redirect vulnerability

MottiniMauro:
An open redirect can be possible when users are able to set the value of
session[:return_to]. If the value used for return_to contains multiple
leading slashes (/////example.com) the user ends up being redirected the
external domain that comes after the slashes (http://example.com).

To fix this issue, extra sanitization was added when processing the
return_to url, removing multiple leading slashes to avoid the open
redirect.

### DIFF
--- a/lib/clearance/authorization.rb
+++ b/lib/clearance/authorization.rb
@@ -86,8 +86,14 @@ module Clearance
     def return_to
       if return_to_url
         uri = URI.parse(return_to_url)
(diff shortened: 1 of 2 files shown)


mike-burns (approved):
Thanks for taking this on.

gnfisher (approved):
:t-rex:

Kirill89 (approved):
👍

PR #950, opened 2021-10-30
Documentation - Update SignInGuard information

matt17r:
As per #891, sign in guards should be specified as strings since Rails 6 and the move to Zeitwerk (update comment in `lib/clearance/sign_in_guard.rb`.)

Additionally, guards in the suggested location (`lib/guards`) don't auto-load for me in Rails 7.0.0.alpha2 (didn't check earlier versions). I propose changing the location in the same file and an addition to the README to help newbies figure out a reasonable folder to build a guardhouse.

Related, [Tute's blog post](https://thoughtbot.com/blog/email-confirmation-with-clearance) could do with a revision now too:

```
  Clearance.configure do |config|
    config.routes = false
-   config.sign_in_guards = [ConfirmedUserGuard]
+   config.sign_in_guards = ["ConfirmedUserGuard"]
  end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ end
 ```
 
 ```ruby
+# app/guards/email_confirmation_guard.rb (or somehwere else it will be auto-loaded)
 class EmailConfirmationGuard < Clearance::SignInGuard
   def call
     if unconfirmed?
--- a/README.md
+++ b/README.md
@@ -349,7 +349,7 @@ end
 ```
 
 ```ruby
-# app/guards/email_confirmation_guard.rb (or somehwere else it will be auto-loaded)
+# app/guards/email_confirmation_guard.rb
 class EmailConfirmationGuard < Clearance::SignInGuard
   def call
     if unconfirmed?
--- a/lib/clearance/sign_in_guard.rb
+++ b/lib/clearance/sign_in_guard.rb
@@ -19,7 +19,7 @@ module Clearance
   #       config.sign_in_guards = ["ConfirmationGuard"]
   #     end
   #
-  #     # in app/guards/confirmation_guard.rb (or another path that is auto-loaded)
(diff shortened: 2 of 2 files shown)


lib/clearance/sign_in_guard.rb
```diff
@@ -16,10 +16,10 @@ module Clearance
   #
   #     # in config/initializers/clearance.rb
   #     Clearance.configure do |config|
-  #       config.sign_in_guards = [ConfirmationGuard]
+  #       config.sign_in_guards = ["ConfirmationGuard"]
   #     end
   #
-  #     # in lib/guards/confirmation_guard.rb
+  #     # in app/guards/confirmation_guard.rb (or another auto-loaded path)
```
MottiniMauro, line 22:
```suggestion
  #     # in app/guards/confirmation_guard.rb
```
I think just updating the path here would be enough to indicate what we want.

README.md
```diff
@@ -349,6 +349,7 @@ end
 ```
 
 ```ruby
+# app/guards/email_confirmation_guard.rb (or somehwere else it will be auto-loaded)
```
MottiniMauro, line 352:
```suggestion
# app/guards/email_confirmation_guard.rb
```
What do you think of just having the path here? I don't think we need to overexplain how rails handles autoload. There is also a typo in `somehwere `

MottiniMauro:
@matt17r Thanks for your contribution! I left a couple of small suggestions for the changes, let me know what you think.

xanderificnl:
I was following along [Tute's post](https://thoughtbot.com/blog/email-confirmation-with-clearance) with Rails 7.

The issue I ran into was `config.sign_in_guards = [ConfirmedUserGuard]` the fix, as mentioned in this PR, is using a string. Perhaps the blog post can also be updated to reflect that?

PR #952, opened 2021-11-19
Add config option to disable signin after password reset

### LINKED ISSUE #949: Disable auto-login after password reset
Currently, users are automatically logged in after resetting a password. This is definitely a smooth UX in most cases. In our application, we’re extending Clearance with MFA support. Therefore, we’d like users to explicitly sign-in after resetting their password.

Currently, this requires overwriting the `PasswordsController#update` action. I’d prefer not having to do that for such a small change and was wondering if you think either a configuration option or extracting that behavior into a separate method that can easily be overwritten might make sense. I’d be willing to put in a PR! :) 

https://github.com/thoughtbot/clearance/blob/main/app/controllers/clearance/passwords_controller.rb#L36

tillprochaska:
Right now, users are automatically signed in after resetting their password. However, for some use cases, it might make sense that they have to sign in explicitly.

For example, in an app we’re building, we have implemented 2FA on top of Clearance. After a password reset, we want users to sign in themselves using the second factor. Otherwise, password resets could be used to circumvent 2FA. (While we could ask for the second factor before resetting the password, we decided to allow password resets without the second factor, but then force users to sign in.)

This PR adds a new configuration option to Clearance, `sign_in_on_password_reset`. It defaults to `true`, i.e. the default behavior does not change.

Closes #949.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Clearance.configure do |config|
   config.sign_in_guards = []
   config.user_model = "User"
   config.parent_controller = "ApplicationController"
+  config.sign_in_on_password_reset = false
 end
 ```
 
--- a/lib/clearance/configuration.rb
+++ b/lib/clearance/configuration.rb
@@ -118,6 +118,12 @@ module Clearance
     # @return [Array<String>]
     attr_accessor :allowed_backdoor_environments
 
+    # Controls wether users are automatically signed in after successfully
+    # resetting their password.
+    # Defaults to `true`.
+    # @return [Boolean]
+    attr_writer :sign_in_on_password_reset
+
     def initialize
       @allow_sign_up = true
       @allowed_backdoor_environments = ["test", "ci", "development"]
@@ -134,6 +140,7 @@ module Clearance
       @secure_cookie = false
       @signed_cookie = false
       @sign_in_guards = []
+      @sign_in_on_password_reset = true
     end
 
     def signed_cookie=(value)
@@ -213,6 +220,10 @@ module Clearance
 
     def rotate_csrf_on_sign_in?
       !!rotate_csrf_on_sign_in
+    end
+
+    def sign_in_on_password_reset?
+      @sign_in_on_password_reset
     end
   end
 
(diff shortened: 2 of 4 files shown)


spec/requests/password_maintenance_spec.rb
```diff
@@ -2,17 +2,36 @@
...
+        expect(cookies["remember_token"]).to be_present
+      end
+    end
+
+    context "if `sign_in_on_password_reset` option is false" do
+      let(:sign_in_on_password_reset) { false }
 
-      expect(response).to redirect_to(Clearance.configuration.redirect_url)
-      expect(cookies["remember_token"]).to be_present
+      it "redirects, but does not sign in the user" do
```
tillprochaska, line 31: I wasn’t sure wether this should go into the controller spec for `PasswordsController` or into this request spec, but decided to add it to the request spec as this one tests the current behavior.

README.md
```diff
@@ -66,6 +66,7 @@ Clearance.configure do |config|
   config.sign_in_guards = []
   config.user_model = "User"
   config.parent_controller = "ApplicationController"
+  config.sign_in_on_password_reset = false
```
tillprochaska, line 69: I think this should be sufficient as the configuration option is quiet self-explanatory and there is additional information in the respective `lib/clerance/configuration.rb`.

lib/clearance/configuration.rb
```diff
@@ -134,6 +140,7 @@ def initialize
       @secure_cookie = false
       @signed_cookie = false
       @sign_in_guards = []
+      @sign_in_on_password_reset = true
```
tillprochaska, line 143: Is there any preference regarding the order of instance variables and method definitions in this class? I simply added the new configuration option at the end. They seem to be grouped somewhat semantically.

dorianmariefr:
I've made some changes and credited you in #969

I think it's better for now to stick to controller specs. Migrating to requests specs would be nice though but the tests are already organized around controller specs.

tillprochaska:
@dorianmariefr That's great, thanks a lot!

PR #955, opened 2022-01-16
Bump Ruby patch versions on GitHub Actions

neilvilela:
Using the latest Ruby patch version for each release gives a better visibility about the Rails (and other gems) compatibility for each of them, as deprecation warnings are added when new minor or major versions are planned.


dorianmariefr:
Fixed with #964

PR #958, opened 2022-03-17
Replace text_field with email_field

stevepolitodesign:
This small change replaces `form.text_field :email, type: 'email'` with `form.email_field :email` in all of the views used in the generators.


stevepolitodesign:
Thank you @dorianmariefr!

PR #963, opened 2022-06-01
Fix return type in documentation

hasghari:
The `email_optional?` and `password_optional?` methods can be overridden to return true so the return type should not be hard-coded to `false`.


dorianmariefr:
I see https://www.rubydoc.info/gems/clearance/2.5.0/Clearance%2FUser:email_optional%3F and https://www.rubydoc.info/gems/clearance/2.5.0/Clearance%2FUser:password_optional%3F

I agree with you, the documentation should indicate the true could be returned

PR #966, opened 2022-06-12
Add unique constraints on remember_token and confirmation_token

dorianmariefr:
This is useful to prevent eventual collisions, possibly when the
application is not using a great random string generator.

For reference Clearance random string implementation is:

```ruby
SecureRandom.hex(20).encode('UTF-8')
```

Adding unique indexes prevents eventual collisions.

This is for new applications, existing applications can still have
duplicate tokens (they will be scoped by user_id during the lookup).

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -23,6 +23,6 @@ ActiveRecord::Schema.define(version: 20110111224543) do
   end
 
   add_index "users", ["email"], name: "index_users_on_email"
-  add_index "users", ["remember_token"], name: "index_users_on_remember_token"
-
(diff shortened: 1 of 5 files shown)


aripollak:
For my understanding about why I should add the unique index, why was this added for confirmation_token? It looks like confirmation_token is only looked up in combination with the user id: https://github.com/thoughtbot/clearance/blob/c8d561cc06e0305613954e48f1ddd445ca359296/app/controllers/clearance/passwords_controller.rb#L59-L60

aripollak:
Oops, nevermind, I see the referenced issue now.

PR #976, opened 2022-07-13
fix validating email in strict mode

manuelmeurer:
To enable strict mode in email validation, [`mode: :strict` should be used](https://github.com/K-and-R/email_validator#strict-mode), not `strict_mode: true`. Maybe it used to be `strict_mode: true` with an earlier version of the `email_validator` gem?
Be that as it is, it seems that the `strict_mode` parameter is ignored and `loose` validation is used.

Now, with strict mode turned on, emails like foo;@example.com, foo@.example.com or foo@example..com are no longer valid, so I removed the specs. Why were they expected to be valid until now? I'm not an expert in that matter, but they look pretty invalid to me. 😸

### DIFF
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -150,7 +150,7 @@ module Clearance
 
       included do
         validates :email,
-          email: { strict_mode: true },
+          email: { mode: :strict },
           presence: true,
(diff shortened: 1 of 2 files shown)


sej3506 (approved):
Thank you!

PR #977, opened 2022-07-26
Update README.md

foucist:
Recommend checking against `Rails.configuration.force_ssl` boolean value in the example configuration.  That way it is off in development & test environment by default.


sej3506 (approved):
Thank you!

Issue #542, opened 2015-03-19
Offer different user_params method for users#create and users#new


derekprior:
Right now both of these actions call `user_from_params`, which calls `user_params`. I think this makes life difficult for people using strong parameters as something they attempt to do with the params for create might cause an error in the new action.

I wonder if there's something quick and backward compatible we can do here...

pedrosmmoreira:
I'd like to help with this, but I'm not sure if I understand the full extent of the issue.

Would it suffice to make `user_params` check the `action` key and then call out to different methods? This would at least be quick and backward compatible.

derekprior:
@pedrosmmoreira This came out of this Stack Overflow question: http://stackoverflow.com/questions/29136246/rails-4-and-thoughtbot-clearance-adding-fields-to-the-user-model/29146029

Consider a case where you expect to have different parameter rules for the user you construct for the `new` action and the user you construct in `create`. Both of these actions call `user_from_params` with no arugments. That method gets the params from `user_params`, again with no params. So there's no obvious way for `user_params` to know if it should be using params for the new action or params for the create action.

Using the action key is one way to do it, but I don't think I want to do that. Feels wrong. Not sure what right is though... any other ideas?

pedrosmmoreira:
Ah, I see. Whenever I've had to deal with this, I've used the same mechanism that is in place: deleting the param from the hash and assigning it to a local var that can get used in the block. For adding a couple of attrs, that doesn't get too bad.

If I am customizing beyond a couple of attrs, I resort to a form object, since at that point it is quite likely I am not pushing everything onto the user anyway. 

I do agree that checking the key is "smelly", but other than some kind of 'Registration' object, I'd go with documenting that 'user_from_params' is the place to extend behaviour. I could also see the initializer as a place for whitelisting, in order to keep compatibility.

j-dexx:
@derekprior Using rails 5 this breaks completely for me with a forbidden attributes error.

derekprior:
@j-dexx What breaks completely? Sign up?

j-dexx:
@derekprior Sorry, that was a pretty unhelpful comment.  I can't find where I had an issue and having created a simple app to try and replicate it I can't.

DavidVII:
[...]

PR #979, opened 2022-08-09
use Strong Parameters on User model user_params method

lipedjow:
This PR suggests 2 small improvements:
- always return ActionParam object on `UsersController` `user_params` method (even if empty)
- use Rails Strong Parameters to permit default user params instead of deleting params entries

this way we can easily override `user_params` method to add new custom fields for User model

let me know if it makes sense to add a configuration entry for changing default new User parameters, for example:
as default: `Clearance.configuration.permit_user_parameters = [:email, :password]`

we could customize it:
`Clearance.configuration.permit_user_parameters = [:name, :email, :password]`

and default permit call would looks like:
`params.fetch(Clearance.configuration.user_parameter, {}).permit(*Clearance.configuration.permit_user_parameters)`

### DIFF
--- a/app/controllers/clearance/users_controller.rb
+++ b/app/controllers/clearance/users_controller.rb
@@ -31,16 +31,10 @@ class Clearance::UsersController < Clearance::BaseController
   end
 
   def user_from_params
-    email = user_params.delete(:email)
-    password = user_params.delete(:password)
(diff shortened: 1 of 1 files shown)


lipedjow:
ref #542

PR #980, opened 2022-09-22
Only overwrite QUERY_STRING if as param present

danielnolan:
* This middleware deletes the as param from the query params if present
  to find the user to sign in. We should only overwrite
  env["QUERY_STRING"] if the as param is present.
* Move line that overwrites env["QUERY_STRING"] inside the
  if user_param.present? so that it will only be overwritten if
  we have deleted the as param.


mike-burns:
Should we `||=` it instead? Is the issue that it's overwriting in general? Should we always overwrite it if the user is present?

mike-burns (approved):
Got it, that makes sense. LGTM!

PR #989, opened 2023-03-08
update argon2 to v2.2.0 due to errors loading argon2_wrap library

schorsch:
happens with rubygems 3.4 and was discussed at: https://github.com/technion/ruby-argon2/issues/56

I just ran `rake spec` locally with ruby 3.1.2p20 and rubygems 3.4.6 and they pass, but i could be missing some more elaborate tests.


brimatteng (approved):
Looks good to me!

PR #995, opened 2023-10-10
Test against rails 7.1

segiddins:

PR #998, opened 2023-12-08
Fix CI - Update workflow, setup, specs, and more

sej3506:
Context:

In #995, a PR to add support for Rails 7.1, CI was failing.
The thing causing this first failure was issue number 1
below. Each additional fix here was yet another thing
CI failed on.

Things in this commit:

1. Split dummy:db:reset to dummy:db:drop and dummy:set:up
2. Replace outdated use of "MiniTest" with "Minitest"
3. Update session_spec.rb `.serialize_cookies` to handle old and
  new versions of Rack's handling of headers
4. Remove deprecated active record handling in application.rb
5. Update request_with_remember_token.rb `remember_token_cookies`
  handling of headers.

---

1. Setup & Github Workflow - Split dummy:db:reset to
  dummy:db:drop and dummy:set:up

This issue was discovered in #995. 

I'm still looking into why exactly this is happening, but solution has
been consistent for me.

To reproduce the issue for yourself:
When CI runs for each version, it does so using
the specific Gemfile for the version.
Running
`BUNDLE_GEMFILE=gemfiles/rails_7.1.gemfile RAILS_ENV=test
bundle exec rake dummy:db:reset`
and then
`bundle exec appraisal rake`
the Users table will not be found, because it is not created from this
command.

This issue only exists with the 7.1 gemfile that I've found.

Splitting the command dummy:db:reset to its component parts of dummy:db:drop and
dummy:db:setup on separate lines works - for all version. Using them on a
single line does not work, though.

2. Replace outdated use of "MiniTest" with "Minitest"

Use of "MiniTest" (vs modern "Minitest") was removed in version 5.19.0.
CI runs using version 5.20 of Minitest and was failing.

https://my.diffend.io/gems/minitest/5.18.1/5.19.0

3. Update session_spec.rb `.serialize_cookies` to handle old and
new versions of Rack's handling of headers

The way `Rack::Utils.set_cookie_header!` works in rack 3.0.8
is different than previous versions.
Rack downcases keys at the class level now. This update supports both
versions of the method.

https://github.com/rack/rack/commit/95d2f64a84a6010d5b1a5179a647b42cb53356d1

Also updating spec/support/cookies.rb and
spec/support/request_with_remember_token.rb for this.

4. Remove deprecated active record handling in application.rb

Rails 6.0 no longer uses `config.activerecord.sqlite3.represent_boolean_as_integer`
as it is always true now.
https://github.com/rails/rails/commit/f59b08119bc0c01a00561d38279b124abc82561b

As well, `config.active_record.legacy_connection_handling` has been
deprecated since Rails 7.0 and throws errors if used.

https://github.com/rails/rails/pull/45835

5. Update request_with_remember_token.rb `remember_token_cookies`
handling of headers.

The value might be a string or an array.


sej3506:
I'll be squashing and merging, using the PR description as the final commit message.

MatheusRich (approved):

mike-burns (approved):

PR #1002, opened 2024-01-14
Add configuration options for failure method redirects

drsharp:
This PR introduces two new accessor values in the Configuration class:
  * `url_after_destroy`
  * `url_after_denied_access_when_signed_out`

These both default to `nil` and their usage defaults to the prior value of `sign_in_url` so that this will be nicely backwards-compatible.

### Rationale
I started a new project and found myself subclassing the `sessions_controller.rb` only so I can override `url_after_destroy`. And it seemed odd that `url_after_create` uses a configuration value but `url_after_destroy` does not.

https://github.com/thoughtbot/clearance/blob/630f1752f4d6e9a186dc511cccd5a5819a5f3077/app/controllers/clearance/sessions_controller.rb#L35-L41

I did the same for `url_after_denied_access_when_signed_out` for completeness' sake.

Note that both of these places will still default to `sign_in_url` if nothing is set. Also, I didn't see any direct unit specs for the authorization module or even the `url_after_denied_access_when_signed_out`

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ Clearance.configure do |config|
   config.mailer_sender = "reply@example.com"
   config.password_strategy = Clearance::PasswordStrategies::BCrypt
   config.redirect_url = "/"
+  config.url_after_destroy = nil
+  config.url_after_denied_access_when_signed_out = nil
   config.rotate_csrf_on_sign_in = true
   config.same_site = nil
   config.secure_cookie = false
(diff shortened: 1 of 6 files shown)


drsharp:
I don't know how important it is to shorten the line lengths that hound is complaining about. The variable is a very long variable but it's also very explicitly named (which I affirm). I could multi-line some of these but they'd be less readable to me. 🤷

sej3506 (approved):
Thank you! This looks good. We can ignore these Hound issues.

PR #1004, opened 2024-03-14
Call dynamic-readme reusable workflow

stefannibrasil:
We want to have a way to edit our README footer at one place and have
the changes from there be propagated to our repos.

By adding this snippet in the README, we call this reusable workflow:
https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml
that renders and updates the README footer dynamically.

This workflow has been updated to accommodate repositories that have
main branch rules. To generate the README dynamically, you’ll need to
merge a PR with the README changes. That will trigger the GitHub workflow.
The PR won’t be generated if the footer does not have any differences.
It will only be open if there’s an update to the footer snippet.

### DIFF
--- a/.github/workflows/dynamic-readme.yml
+++ b/.github/workflows/dynamic-readme.yml
@@ -1,0 +1,19 @@
+name: update-templates
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - README.md
+  workflow_dispatch:
+
+jobs:
+  update-templates:
+    permissions:
+      contents: write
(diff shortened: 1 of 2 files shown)


stefannibrasil:
@sej3506 any issues into merging this one? No problem if you'd prefer to wait for https://github.com/thoughtbot/clearance/pull/1005 to be merged first. I went ahead and rebased that one for you, let me know if I missed anything ⭐

PR #1005, opened 2024-03-18
Drop Ruby 2.76 and Rails 6.0

sej3506:
- Drops support for Rails 6.0, see https://endoflife.date/rails
- Drops support for Ruby 2.7.6, see https://endoflife.date/ruby


stefannibrasil (approved):
:shipit:

PR #1006, opened 2024-03-18
Replace APOSTROPHE with RIGHT SINGLE QUOTATION MARK in locale

sej3506:
Replace APOSTROPHE (U+0027) with RIGHT SINGLE QUOTATION MARK (U+2019) in locale.
See https://github.com/rails/rails/pull/45463

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ complete changelog, see the git history for each version via the version links.
 
 ## [Unreleased]
 
+- Replace APOSTROPHE (U+0027) with RIGHT SINGLE QUOTATION MARK (U+2019) in locale.
+See https://github.com/rails/rails/pull/45463
+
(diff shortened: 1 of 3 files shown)


stefannibrasil (approved):
Thank you!

dorianmariecom:
I don't think it's a good idea, knowing that the change was reverted: https://github.com/rails/rails/commit/acfa0454058c535c1352ddab91c93ca914f0805a

> Revert typography change in user facing errors
> This change would force a lot of existing applications and libraries
> to update their tests.
> 
> We included it in the beta to collect feedback from the community and
> we had some comments about how negative this change would be.
> 
> Developers that care about the typography of their error messages
> can easily change it in their applications using the translation
> files, so there is no need to inflict pain in the upgrade process
> by changing the default in the framework.

PR #1009, opened 2024-03-18
Commit appraisals gemfile locks

stefannibrasil:
I don't know why the gemfile locks from appraisals weren't being included.

But they are helpful to keep - just like the main Gemfile.

I don't see why not committing them. Open to be convinced of the contrary though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 *~
 .bundle
 db/*.sqlite3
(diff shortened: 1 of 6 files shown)


stefannibrasil:
If this one gets merged, we don't need to poke with CI to verify the issue is with our appraisals not having been updated with the latest railties versions @sej3506

PR #1015, opened 2024-03-29
Update specs to match on translations

sej3506:
In Rails 7.13.1 and 7.0.8.1, a XSS vulnerability was closed that updated how translated strings are dealth with.
https://github.com/advisories/GHSA-9822-6m93-xqf4

This change to specs updates how we verify acceptance (using the translation instead of a regex) and also adds in a helper to conditionally html escape the string based on the Rails version.


sej3506:
Failing in 2.7.6 which will have support dropped just as soon as I merge in this fix.

Issue #993, opened 2023-07-21
Feature request: Disable public password reset.


a-chacon:
Hello! I am using this gem, and it is excellent and simple. I was looking for some configuration to disable the password reset function, I am using this for an admin panel login and I don't want a reset password, maybe it can be disabled by default and only enable when you want to use it. Is it possible to have this configuration? If this has some sense for you, I can try to do it. Only if you think this can be helpful.

For now, I can remove the link in the sign-in view, but the route is still available and public.

I apologize if there is a solution for this that I didn't see.

sej3506:
Added in #1019

PR #1019, opened 2024-05-08
Add allow_password_resets config option

whatnotery:
solves #993 

added config option to disallow password reset routes in cases where its not needed.  `allow_password_resets` defaults to `true` but can be set to `false` in instances where it is not needed.


sej3506 (approved):
This looks great! Thank you for the contribution and the awesome spec coverage!

Issue #999, opened 2023-12-16
Circular require


segiddins:
Run the following script with `RUBYOPT=-w`

```ruby
require "rubygems"
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "rails"
  gem "clearance", require: "clearance"
end
```

Note that the following warnings are printed:

```
[/Users/segiddins/.gem/ruby/3.2.2/gems/clearance-2.6.1/lib/clearance/configuration.rb:198](coderunner:///Users/segiddins/.gem/ruby/3.2.2/gems/clearance-2.6.1/lib/clearance/configuration.rb#198:0!): warning: method redefined; discarding old user_parameter
<internal:/Users/segiddins/.rubies/ruby-3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:38: warning: <internal:/Users/segiddins/.rubies/ruby-3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:38: warning: loading in progress, circular require considered harmful - /Users/segiddins/.gem/ruby/3.2.2/gems/clearance-2.6.1/lib/clearance.rb
	from Untitled 2.rb:7:in  `<main>'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler/inline.rb:42:in  `gemfile'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler.rb:403:in  `with_unbundled_env'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler.rb:658:in  `with_env'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler.rb:403:in  `block in with_unbundled_env'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler/inline.rb:51:in  `block in gemfile'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler/settings.rb:158:in  `temporary'
	from /Users/segiddins/.gem/ruby/3.2.2/gems/bundler-2.5.1/lib/bundler/inline.rb:66:in  `block (2 levels) in gemfile'
[...]

PR #1027, opened 2024-06-20
Fix method redefinition and circular require issues

kholdrex:
https://github.com/thoughtbot/clearance/issues/999

This PR addresses two issues:

1. **Method Redefinition:** The `user_parameter` method was being defined multiple times in `lib/clearance/configuration.rb`. This has been fixed by changing `attr_accessor` to `attr_writer`, ensuring the method is defined only once.

2. **Circular Require:** Circular dependencies between `lib/clearance.rb` and `lib/clearance/engine.rb` have been resolved by rearranging the require statements to avoid circular loading.

### DIFF
--- a/lib/clearance.rb
+++ b/lib/clearance.rb
@@ -5,9 +5,9 @@ require 'clearance/rack_session'
 require 'clearance/back_door'
 require 'clearance/controller'
 require 'clearance/user'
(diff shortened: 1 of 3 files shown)


sej3506 (approved):
Thank you! Good catch with the user_parameter method(s)!

PR #1032, opened 2024-08-13
ci: Add Rails 7.2 and Ruby 3.3 to testing matrix

hasghari:


spec/dummy/application.rb
```diff
@@ -12,6 +12,11 @@ class Application < Rails::Application
     if Rails.version.match?(/(6.1|7.0)/)
       config.active_record.legacy_connection_handling = false
     end
+
+    if Rails.gem_version >= Gem::Version.new("7.1")
```
hasghari, line 16: Ideally I would like to add `config.load_defaults Rails::VERSION::STRING.to_f` to this file and remove these conditional settings that only exist to address the deprecation warnings. However, when loading the defaults, the specs in the `spec/requests` directory fail.

.gitignore
```diff
@@ -5,7 +5,6 @@
 *~
 .bundle
 db/*.sqlite3
-gemfiles/*.lock
```
hasghari, line 8: I was seeing failures on my local system when running the specs that weren't happening on GitHub. I tracked this down to different versions of gems being installed locally versus CI so it makes sense to commit the lockfiles in order to achieve parity between local and CI.

lib/clearance/back_door.rb
```diff
@@ -48,11 +48,13 @@ def call(env)
...
     def sign_in_through_the_back_door(env)
-      params = Rack::Utils.parse_query(env["QUERY_STRING"])
+      params = Rack::Utils.parse_query(env[Rack::QUERY_STRING])
       user_param = params.delete("as")
 
       if user_param.present?
-        env["QUERY_STRING"] = Rack::Utils.build_query(params)
+        query_string = Rack::Utils.build_query(params)
+        env[Rack::QUERY_STRING] = query_string
+        env[Rack::RACK_REQUEST_QUERY_STRING] = query_string
```
hasghari, line 57: This fixes the following deprecation warning:
```
warning: query string used for GET parsing different from current query string. Starting in Rack 3.2, Rack will used the cached GET value instead of parsing the current query string.
```

sej3506:
There are a number of files deleted in this PR. Could you add a description and some information about what this PR is doing beyond updating the test matrix?

hasghari:
> There are a number of files deleted in this PR. Could you add a description and some information about what this PR is doing beyond updating the test matrix?

@sej3506 Sure. I've tried to have descriptive commit messages for each of the 6 commits in this PR. However, the commit with the subject "Use default settings of Rails version under test" is the most substantial one in this PR. Many of the deprecation warnings that were produced were a result of testing different versions of Rails with non-default settings. My goal with that specific commit was to set `config.load_defaults Rails::VERSION::STRING.to_f` so that the default settings would be applied for testing the specific version of Rails under test.

I also wanted to standardize the setup of a dummy application for testing purposes so I generated the dummy application structure by copying the files generated by running `rails plugin new`. This ends up simplifying the configuration and removing some of the monkey-patching that was necessary in the old setup.

sej3506 (approved):
Thank you for these upgrades! I think a lot of headaches have come from the dummy application. I appreciate the straight forward and modern update to it!

PR #1041, opened 2024-11-14
Fix query params being clobbered by Clearance::BackDoor

### LINKED ISSUE #1040: Query params removed after update from version 2.8.0 to 2.9.1
**Description:**

After updating the gem from version 2.8.0 to 2.9.1, I encountered an issue where certain query parameters are no longer passed correctly. Specifically, the date parameter and other query parameters are missing from the params hash in GET requests, although they were present in version 2.8.0.

**Steps to Reproduce:**

1. Make a GET request to some_url with query parameters, using version 2.8.0:
```ruby
get some_url(something.id, date:, format: "csv", as: admin_user)
```

2. Check params.keys. The output is as expected:
```ruby
> params.keys
=> ["date", "controller", "action", "something_id", "format"]
```

3. Update the gem to version 2.9.1.
4. Make the same GET request with the identical parameters:
```ruby
get some_url(something.id, date:, format: "csv", as: admin_user)
```
5. Check params.keys again. This time, the date parameter (and all other query parameters) are missing:
```ruby
> params.keys
=> ["controller", "action", "something_id", "format"]
```

**Expected Behavior:**

The date and other query parameters should still be included in params.keys after updating to version 2.9.1.

**Actual Behavior:**

The date parameter (along with other query parameters) is no longer present in params.keys.

**Environment:**

Gem version: 2.9.1 (works as expected in 2.8.0)
Ruby version: 3.3.5
Rails version: 7.2.1.2 

**Additional Information:**
- I tested with version 2.9.0, and the issue seems to have been introduced there. Version 2.9.0 behaves the same as version 2.9.1, where the query parameters are missing, while version 2.8.0 works as expected.
- I tried passing other query parameters, and all of them are removed from params.keys in versions 2.9.0 and 2.9.1.

Has anyone else experienced this issue?

fcheung:
In rack 3.1.x Setting Rack::RACK_REQUEST_QUERY_STRING causes rack to think
that the query string has already been parsed
(see https://github.com/rack/rack/blob/v3.1.7/lib/rack/request.rb#L487)

This was introduced in #2703 but wasn't actually necessary - the warning mentioned
in that PR is only triggered if only Rack::RACK_REQUEST_QUERY_STRING is updated,
but the correct behaviour is to only set Rack::QUERY_STRING, not to set both

Fixes #1040

### DIFF
--- a/lib/clearance/back_door.rb
+++ b/lib/clearance/back_door.rb
@@ -54,7 +54,6 @@ module Clearance
       if user_param.present?
         query_string = Rack::Utils.build_query(params)
         env[Rack::QUERY_STRING] = query_string
-        env[Rack::RACK_REQUEST_QUERY_STRING] = query_string
         user = find_user(user_param)
         env[:clearance].sign_in(user)
       end
--- a/spec/dummy/app/controllers/application_controller.rb
+++ b/spec/dummy/app/controllers/application_controller.rb
@@ -2,6 +2,6 @@ class ApplicationController < ActionController::Base
   include Clearance::Controller
 
   def show
-    render inline: "Hello user #<%= current_user.id %>", layout: false
+    render inline: "Hello user #<%= current_user.id %> #{params.to_json}", layout: false
   end
 end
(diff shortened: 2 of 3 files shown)


sej3506:
Awesome! Thank you! It was fun watching your process solving this in person at RubyConf. 😁

PR #1044, opened 2025-02-27
normalize blank emails to nil

manuelmeurer:
I noticed that `Clearance::User.normalize_email` will convert nil to an empty string, since it always calls `to_s` on it.
I think normalizing an email that is blank should convert it to nil.

In my app, I want to allow empty emails but still validate their uniqueness, if an email is given. This only works when empty emails are stored as nil, not as an empty string, which is the current behavior.

### DIFF
--- a/lib/clearance/user.rb
+++ b/lib/clearance/user.rb
@@ -127,6 +127,7 @@ module Clearance
       end
 
       def normalize_email(email)
+        return if email.blank?
         email.to_s.downcase.gsub(/\s+/, "")
       end
 
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -75,7 +75,9 @@ describe User do
 
       expect(user_does_not_exist_time). to be_within(0.01).of(user_exists_time)
     end
+  end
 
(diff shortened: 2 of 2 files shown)


manuelmeurer:
In the second commit, I moved a test out of a `describe` block to which it was not related.

manuelmeurer:
Test failures don't seem to have anything to do with my change...

sej3506:
I don't think we will add this change since it could have negative effects for many gem users that expect the empty string. I would consider converting nil to "" to fall in line with the concept of normalizing -- every input should come out that same type.

manuelmeurer:
> I don't think we will add this change since it could have negative effects for many gem users that expect the empty string.

Sure, it would be a breaking change, but why not add it in the next major version?

> I would consider converting nil to "" to fall in line with the concept of normalizing -- every input should come out that same type.

I don't quite understand...

The current behavior converts nil to an empty string, with my changes nil will stay nil, and an empty string will be nil as well. This is what "normalizing" means IMO.
Would you accept nil -> nil and empty -> empty?

PR #1045, opened 2025-03-28
Drop support for Rails 7.0 and Ruby 3.1

sej3506:
Drops support for Rails 7.0, see https://endoflife.date/rails
Drops support for Ruby 3.1, see https://endoflife.date/ruby

Also removing `HTMLEscapeHelper` which was a test utility I
added just to support html escaping strings in tests for Rails 7.0.


.github/workflows/tests.yml
```diff
@@ -16,17 +16,12 @@ jobs:
...
           - "8.0"
         ruby:
-          - "3.1.6"
-          - "3.2.5"
-          - "3.3.5"
-        exclude:
-          - ruby: "3.1.6"
-            gemfile: "8.0"
+          - "3.2.8"
+          - "3.3.7"
```
MatheusRich, line 24: separate issue, but we need to test on 3.4.2

MatheusRich (approved):
LGTM.

MatheusRich:
@sej3506 are you planning on cutting a new release? Should we add a CHANGELOG entry?

brian-penguin (approved):

sej3506:
> @sej3506 are you planning on cutting a new release? Should we add a CHANGELOG entry?

Yes, I plan to do that in a separate commit 👍🏻

.github/workflows/tests.yml
sej3506: Sigh, I GUESS 😆

PR #1085, opened 2026-03-02
Replace obsolete unprocessable_entity with unprocessable_content

sej3506:
This change replaces all instances of :unprocessable_entity with :unprocessable_content.

In keeping aligned with the IANA HTTP Status Code Registry, Rack deprecated unprocessable_entity for unprocessable_content. They still include handling for the time being but have included a deprecation notice.

More info: https://github.com/rack/rack/pull/2137

### DIFF
--- a/app/controllers/clearance/passwords_controller.rb
+++ b/app/controllers/clearance/passwords_controller.rb
@@ -38,7 +38,7 @@ class Clearance::PasswordsController < Clearance::BaseController
       session[:password_reset_token] = nil
     else
       flash_failure_after_update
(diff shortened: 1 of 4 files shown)


FerPerales (approved):
Thank you! @sej3506 

Do we need to upgrade rake/rails as well? 🤔

sej3506:
@FerPerales We wouldn't need to at this time. The changes have been live in rack since at least 2024 and we've been getting the deprication warning in our tests on CI, too. [Example](https://github.com/thoughtbot/clearance/actions/runs/19935236101/job/57158038710#step:5:10).

PR #1086, opened 2026-03-12
Add configurable password reset token expiration

aripollak:
I took a stab at implementing thoughtbot/clearance#465.

Introduces `password_reset_token_expiration_in` configuration option (defaults to nil, preserving existing no-expiry behavior). When set, `forgot_password!` stamps `confirmation_token_created_at` and `password_reset_token_expired?` enforces the window.

The controller rejects expired tokens the same way it rejects invalid ones. The generator and migration templates add the new column so existing installs get it via the standard `rails generate clearance:install` additive migration path.

Guards against the column not yet existing (pre-migration) so the feature is safe to deploy before running migrations.


FerPerales:
@aripollak Thanks for the PR. Looks promising! I'll do a proper code review next week and provide feedback.

FerPerales:
@sej3506 you may also want to take a look :)

lib/clearance/configuration.rb
```diff
@@ -69,6 +69,18 @@ class Configuration
...
 
+    # How long a password reset token is valid. Defaults to `nil` (no
+    # expiration). Set to an ActiveSupport::Duration to enable expiration, e.g.
+    #
+    #   config.password_reset_token_expiration_in = 2.hours
+    #
+    # When set, `forgot_password!` records the time the token was issued.
+    # Tokens with no recorded issue time (e.g. issued before upgrading or
+    # before running the migration) are treated as expired, forcing users to
+    # re-request a reset.
```
aripollak, line 80: I'm guessing this is too much explanation for a configuration variable.

FerPerales:
@aripollak this looks good! We have a couple of upgrades/security releases we'd like to do first and then we'll ship this. Thank you!

PR #1088, opened 2026-04-03
Upgrade Ruby and Rails versions

FerPerales:
* Remove Ruby 3.2.9
* Bump ruby 3.3.9 to 3.3.11
* Make ruby 3.3.11 the minimum required version
* Bump ruby 3.4.5 to 3.4.9
* Add ruby 4.0.2
* Remove RoR 7.1.0 support
* Add RoR 8.1.0 support

### DIFF
--- a/.github/workflows/standardrb.yml
+++ b/.github/workflows/standardrb.yml
@@ -15,5 +15,5 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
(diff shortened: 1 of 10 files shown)


jaredlt:
@FerPerales what's the thinking with requiring such a new minimum ruby version? Ruby 3.3.11 was released 2026-03-26

I expect that might be a sticking point for some folks to be able to upgrade. Do we use language features that aren't on older ruby versions?

